PR #27, opened 2026-02-09
feat: account functionality

juliansommer:


juliansommer:
deferring for now. need console apis + viper block encryption support for this to actually work. current code just puts it in jetstream and we do nothing with it on volume creation

PR #139, opened 2026-04-14
Feat/iso installer

tomnewton-mulga:
Working iso installer - contains some fluff for testing setup (tofu-iso) that will be better implemented as extra envs in tofu-cluster when Ben gets around to it


build/scripts/wait-for-nats.sh
```diff
@@ -7,7 +7,7 @@ CONF=/etc/spinifex/spinifex.toml
 NATS_CONF=/etc/spinifex/nats/nats.conf
 NATS_HOST=""
 NATS_PORT="4222"
-TIMEOUT=30
+TIMEOUT=90
```
juliansommer: why was this changed? waiting 90s for nats seems like a lot and if we have to wait that long isn't it better to fail out?

cmd/installer/install/install.go
```diff
@@ -0,0 +1,556 @@
...
+		return os.WriteFile(filepath.Join(udevDir, "70-spinifex-net.rules"), []byte(udevRules.String()), 0o644)
+	}
+	return nil
+}
+
+// writeLANBridgeUnit installs a non-critical oneshot service that brings up
+// br-lan *after* network-online.target. This keeps br-lan out of
+// networking.service entirely — a missing LAN cable or DHCP timeout on the
+// secondary bridge can never stall the management interface or firstboot.
+func writeLANBridgeUnit(root string) error {
```
juliansommer: could we put all these service files into installer/systemd ?

cmd/installer/firstboot/firstboot.go
```diff
@@ -0,0 +1,252 @@
...
+	clusterCmd := buildClusterCmd(cfg)
+
+	// --encap-ip is optional: when DHCP is used the IP is unknown at install time
+	// and setup-ovn.sh auto-detects it from the default route at boot.
+	setupOVN := "/usr/local/bin/setup-ovn.sh --management"
+	if cfg.EncapIP != "" {
+		setupOVN += fmt.Sprintf(" --encap-ip=%s", cfg.EncapIP)
+	}
+
+	script := fmt.Sprintf(`#!/bin/bash
```
juliansommer, line 81: move those scripts into installer/scripts too

juliansommer (dismissed):

PR #182, opened 2026-05-01
fix(vpc): validate SG rule CidrIp/SourceSG to block OVN ACL injection

juliansommer:
## Summary

Implements **Finding 1** of [`spinifex-security-audit-findings.md`] tenant-controlled string injection into OVN ACL match expressions via `CidrIp` / `SourceSG`.

`vpcd/acl.go` builds OVN match expressions with `fmt.Sprintf` from values that flow straight from the AWS request body. OVN's `&&` binds tighter than `||`, so `CidrIp = "1.2.3.4/32 || ip4.src == 0.0.0.0/0"` produces a match equivalent to *"allow all IPv4 ingress"* — silently bypassing port/SG-membership constraints. `SourceSG` is the same primitive and could reference another tenant's port group via `@<other_pg>`.

## Changes

- **Handler validation** (`handlers/ec2/vpc/security_group.go`):
  - `CidrIp` validated with `net.ParseCIDR` + canonical round-trip (rejects host bits set, trailing garbage, anything whose `ipnet.String()` differs from input).
  - `SourceSG` validated against `^sg-[0-9a-f]{17}$` (matches `utils.GenerateResourceID("sg")`).
  - Applied in `AuthorizeSecurityGroupIngress`, `AuthorizeSecurityGroupEgress`, `RevokeSecurityGroupIngress`, `RevokeSecurityGroupEgress` — rejects with `InvalidParameterValue` before persisting or publishing the SG event.
- **Defensive guard** (`services/vpcd/acl.go`):
  - `BuildIngressACLMatch` / `BuildEgressACLMatch` now return `(string, error)` and refuse any `CidrIp` / `SourceSG` containing OVN match-expression separators (`& | ( ) = < > $ @`, whitespace, quotes, control chars).
  - If it ever fires it is a handler-validator bug — `addRuleACLs` logs `slog.Error` and skips installing the unsafe ACL. Default-deny at priority 900 keeps the port group safe in that case.


juliansommer:
merge post release

PR #210, opened 2026-05-11
feat(viperblockd): retry LoadState on transient backend errors (siv-25 medium-term)

joshsiv-mulga:
## Summary

- Wraps `vb.LoadState()` in `viperblockd` mount handler with exponential backoff: 5 attempts, 200ms × 1.5ⁿ (worst-case ~3.7s, well under the 30s `ebs.mount` NATS deadline).
- Retries only on `viperblock.ErrStateBackendUnavailable`. `ErrStateNotFound` and any unclassified error fail fast so real problems stay visible.
- Retry loop is parameterised over `sleep` and `loadFn`, so the policy is unit-tested without spinning up a real VB.

Depends on viperblock PR introducing the sentinels (`feat/siv-25-state-error-classification`, mulgadc/viperblock#11). This is the medium-term half of mulga-siv-25 (short-term shipped in #209: `MarkRecoveryFailed`).

## Test plan
- [x] `make preflight` clean
- [x] `TestRetryLoadState` covers: first-attempt-success / transient-then-recover / not-found-no-retry / unclassified-no-retry / budget-exhausted backoff schedule
- [ ] Manual: `systemctl restart spinifex-daemon` with predastore deliberately slow to start; expect mount to succeed within retry budget instead of falling into recovery cleanup

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,7 +13,7 @@ require (
 	github.com/kdomanski/iso9660 v0.4.0
 	github.com/klauspost/cpuid/v2 v2.3.0
 	github.com/mulgadc/predastore v1.2.1
-	github.com/mulgadc/viperblock v1.2.1
+	github.com/mulgadc/viperblock v1.2.2-0.20260511064619-1aa54f0dc1f6
 	github.com/nats-io/nats-server/v2 v2.14.0
 	github.com/nats-io/nats.go v1.51.0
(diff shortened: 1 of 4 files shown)


joshsiv-mulga:
Re-running CI: viperblock dev squash-merged to main (commit 69f7a47), so the sentinel symbols are now on the default branch the workflow checks out.

PR #255, opened 2026-05-26
feat(network): ADR-0006 layer redesign + IPsec/OVN hardening

joshsiv-mulga:
## Summary

ADR-0006 network-layer redesign: 68 commits restructuring the network stack into the L0–L8 contract, plus IPsec rollout fixes, OVN robustness, and the duplicate-router TOCTOU fix from `mulga-siv-146`.

### Highlights

- **Network layer split (ADR-0006).** Relocate `services/vpcd` → `spinifex/vpcd`, NATS VPC subscribers → `network/subscribers`, `nbdb` model → `network/ovn`, L0 helpers → `network/host`. Pin invariants in `network/invariants/` (ADR-0006 S2/S4/S8) with AST-walk tests.
- **`topology.Port` migration (S4).** Daemon, handlers, VM all consume `topology.Port` for OVN port names; bridge-mode bleed and dead retrofit helpers removed.
- **OVN duplicate-entity TOCTOU fix (`mulga-siv-146`).** New `Ensure{LogicalRouter,LogicalSwitch,PortGroup}` primitives use OVSDB wait-op + insert in a single transaction so concurrent callers converge to a single row. Bare `Create*` blocked outside `network/ovn` by `TestS4_OVNNamedCreatesUseEnsure`.
- **IPsec hardening (S8).** Orchestrator moved out of daemon. Charon AppArmor read grants for `/etc/spinifex/ipsec` and cluster CA. CA chain symlinked into charon's `cacerts` dir + `rereadcacerts` trigger. `openvswitch-ipsec` + `strongswan-charon` installed at provision time.
- **OVN chassis identity.** OVS `system-id` pinned to hostname / `--node-name` so IPsec cert subject and chassis-id agree.
- **Reconcile.** Skip orphan prune on startup (drift owns GC); DHCP_Options keyed by `subnet_id` not CIDR.
- **CI/governance.** `test_run` workflow input for targeted `-test.run` filter. CODEOWNERS gates `network/` + `vpcd/` paths.

### Scope

172 files changed (+14233 / −15621). Net shrink driven by dead-code removal during the vpcd relocation.

## Test plan

- [x] `make preflight` green on tip (`6f59ec2`)
- [x] ADR-0006 invariants pass: `TestS4_OVNNamedCreatesUseEnsure`, port-name suite, host-layer scan
- [x] Ensure* concurrency tests (50/30/30 goroutines for Router/Switch/PortGroup) — single UUID survives
- [x] `TestLiveManager_EnsureVPC` regression
- [x] Multinode E2E `TestMultinodeGuestSSH` rerun on the GHA workflow (failure originally tracked in run 26447541192)
- [ ] Smoke: IPsec handshake on fresh node — verify charon picks up CA via symlink + `rereadcacerts`

## Refs

- ADR-0006 (`docs/development/proposals/az-local-architecture/0006-spinifex-network-layer-contract.md`)
- Bead `mulga-siv-146` — duplicate routers / SSH timeout


cmd/spinifex/cmd/templates/spinifex.toml
juliansommer: bump version number, can skip the migration but bump

spinifex/admin/ipsec.go
```diff
@@ -0,0 +1,247 @@
...
+// ovs-monitor-ipsec generates strongSwan configs with leftid=@<chassis-id>
+// / rightid=@<chassis-id> and charon validates those names against the
+// peer cert's dnsName SAN. setup-ovn.sh pins system-id to `hostname -s`
+// for exactly this reason — see the comment in setup-ovn.sh near
+// `system-id.conf`. Changing the cert identity here without updating
+// setup-ovn.sh (or vice versa) breaks IKE with AUTHENTICATION_FAILED.
+// GenerateIPSecPeerCertIfEnabled is a no-op when enabled is false. Used by
+// admin init and admin join so the call site reads as a single guarded line
+// keyed off the cluster's network.ipsec_enabled flag.
+func GenerateIPSecPeerCertIfEnabled(enabled bool, configDir, hostname, nodeIP string) error {
```
juliansommer: idk what code calls this but could this not be done in the caller instead of seperate function?

spinifex/daemon/daemon.go
```diff
@@ -1013,6 +1014,25 @@ func (d *Daemon) startCluster() error {
 		return fmt.Errorf("initialize JetStream: %w", err)
 	}
 
+	// Tombstone the spinifex-dhcp-leases bucket left over from the upstream-DHCP
+	// client removed in Phase 2.3 (bead mulga-siv-125.3.3). Idempotent — first
+	// daemon start on an upgraded cluster sweeps the bucket; later restarts are
+	// no-ops.
+	if js, jsErr := d.natsConn.JetStream(); jsErr == nil {
```
juliansommer, line 1020: should this be a migration? get claude to understand the existing bucket migration path, it could handle this.

spinifex/handlers/ec2/igw/service_impl.go
```diff
@@ -291,7 +291,7 @@ func (s *IGWServiceImpl) AttachInternetGateway(input *ec2.AttachInternetGatewayI
 	}
 
 	record.VpcId = vpcID
-	record.State = "available"
```
juliansommer, line 294: I'm pretty sure there's some implicit logic that relies on this being available, get claude to scan but I think I had an issue earlier with e2e regarding available vs attached

spinifex/handlers/ec2/vpc/service_impl.go
```diff
@@ -35,13 +35,18 @@ const (
...
+// matching `vpc.AZ == localAZ` (with the legacy rule that empty matches
+// every AZ for pre-Phase-2.2 records). New records always populate AZ from
+// the daemon's config.AZ.
 type VPCRecord struct {
 	VpcId                            string            `json:"vpc_id"`
 	CidrBlock                        string            `json:"cidr_block"`
 	State                            string            `json:"state"`
 	IsDefault                        bool              `json:"is_default"`
 	VNI                              int64             `json:"vni"`
+	AZ                               string            `json:"az,omitempty"`
```
juliansommer, line 49: with these changes to AZ, do we need to change the DescribeAvailabilityZone command? it currently just reads from config?

spinifex/network/host/health.go
```diff
@@ -0,0 +1,48 @@
...
+	status := OVNHealth{}
+
+	if err := utils.SudoCommand("ovs-vsctl", "br-exists", "br-int").Run(); err == nil {
+		status.BrIntExists = true
+	}
+
+	// Trixie OVN moved the pid file to /var/run/ovn/; older installs use
+	// /var/run/openvswitch/. Probe at runtime so the same binary works on both.
+	ovnTarget := "ovn-controller"
+	if _, err := os.Stat("/var/run/ovn/ovn-controller.pid"); err == nil {
```
juliansommer, line 30: might be out of scope for this but can we improve the ovn pid detection? we always run into this, can we detect it at startup or something and store it globally somewhere and just use utils.getOVNPID ?

spinifex/network/host/host.go
```diff
@@ -0,0 +1,115 @@
+// Package host is L0 of the spinifex network stack: kernel/OVS host wiring.
+// It owns the host's OVS bridges (br-int, br-ext), the uplink port that
+// attaches br-ext to the WAN, and the external CIDR readable off the WAN
+// bridge. Bridge mode (physical vs veth) is resolved here and is invisible
+// above L0; layers L1+ receive UplinkMode as an init-time enum.
+//
+// See docs/development/feature/spinifex-network-redesign.md §5 and
```
juliansommer: remove plan ref

spinifex/network/host/ipsec_enable.go
```diff
@@ -0,0 +1,121 @@
...
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/mulgadc/spinifex/spinifex/admin"
+	"github.com/mulgadc/spinifex/spinifex/config"
+	"github.com/mulgadc/spinifex/spinifex/utils"
+)
+
+// systemctlActiveTimeout bounds how long EnableOVNIPSec waits for
```
juliansommer: in general can you get claude to go through and trim comments, some are really excessive.

spinifex/config/config.go
```diff
@@ -244,5 +245,66 @@ func LoadConfig(configPath string) (*ClusterConfig, error) {
...
+		return nil, err
+	}
+
 	return &config, nil
 }
+
+// validateClusterConfig rejects legacy upstream-DHCP config and validates
+// external pool ranges. The upstream-DHCP-client model was replaced by static
+// WAN-pool allocation; legacy keys must be removed from the operator's TOML.
+func validateClusterConfig(cc *ClusterConfig) error {
```
juliansommer, line 258: can we just remove this? assume everyone has working config, if we want to make sure the config isn't broken, create migration?

spinifex/daemon/daemon.go
joshsiv-mulga: There's no path to drop entire buckets, only mutate KVs inside buckets. If we have more deprecated buckets in the future, might be worth adding a migration?

spinifex/config/config.go
joshsiv-mulga: Having external_dhcp in config actually does nothing, it would be ignored. The validation on the IP range needs to occur somewhere as users are still able to set invalid ranges (higher start address than end address, etc). A migration would only remove external_dhcp, which has no effect anyway.

spinifex/network/host/health.go
joshsiv-mulga: probably good to pull it up if this is an issue elsewhere. I will create an issue.

spinifex/daemon/daemon.go
juliansommer: ok, leave for now and in future we can wrap some migration command which clears old buckets.

spinifex/handlers/ec2/vpc/service_impl.go
joshsiv-mulga: Yes this will have to change when we introduce multi-AZ

PR #259, opened 2026-05-27
chore(ci): disable broken DHCP cells in nightly matrix

joshsiv-mulga:
## Summary

- Disable cells 1, 3, 8, 10 (all `net: dhcp`) in `e2e-nightly.yml`. They fail bootstrap because `spx admin init` with no `--external-pool` still defaults to `--external-mode=pool` and then errors out requiring the pool.
- Static cells (2, 9, 17, 18) already cover the same topo/host combinations, so coverage is preserved.
- Add a `disabled: true` field to each DHCP cell and extend the gate step to skip them on scheduled runs. Override is still available via `workflow_dispatch` with an explicit `cells=1,3,8,10` input.

Reference failing run: https://github.com/mulgadc/spinifex/actions/runs/26527014561

Follow-up: fix the zero-config DHCP path in `spx admin init` so pool-less bootstrap works, then drop the `disabled` flags.

## Test plan

- [ ] Trigger `workflow_dispatch` with blank input on this branch → only cells 2, 9, 17, 18 run.
- [ ] Trigger `workflow_dispatch` with `cells=1` → cell 1 runs (confirms override path).
- [ ] Next scheduled nightly run is all-green.

### DIFF
--- a/.github/workflows/e2e-nightly.yml
+++ b/.github/workflows/e2e-nightly.yml
@@ -27,12 +27,16 @@ jobs:
       max-parallel: 3
       matrix:
         include:
-          - { cell: 1,  install: binary, topo: single,     net: dhcp,   host: debian-13,    arch: x86_64, runner: ci-single, extra: "" }
+          # DHCP cells (1, 3, 8, 10) disabled: zero-config DHCP path in `spx admin init`
+          # is broken (requires --external-pool when --external-mode=pool). Static cells
+          # (2, 9, 17, 18) provide equivalent topo/host coverage. Re-enable by removing
+          # the `disabled` field once the init flow supports pool-less DHCP.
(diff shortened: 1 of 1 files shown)


juliansommer:
why are we disabling it? should we not just fix the dhcp issue?

joshsiv-mulga:
> why are we disabling it? should we not just fix the dhcp issue?

EIPs are no longer assignable by upstream router DHCP - a static pool must be assigned. This aligns with AWS.

Internal DHCP manager was removed, so the only DHCP from an upstream router possible is for a node's WAN address.

Need to update the docs.

juliansommer:
I just realised the cron job goes off what workflow is on main, so would need to cherry pick this and then merging dev to main on Monday is painful, can we just leave it and let the dhcp cells fail for now?

joshsiv-mulga:
Cancelling this PR - will reintroduce internal DHCP manager for EIP

PR #260, opened 2026-05-28
Feat/mig capable

tomnewton-mulga:
Added MIG capability for GPU passthrough and docs from SMCI and Arcee Trinity workloads


juliansommer:
fix the lint issues, and diff coverage (if possible, don't add crappy tests) and LGTM

PR #334, opened 2026-06-10
fix: resolve EIP NAT MAC from OVN LSP so all floating IPs survive reboot

joshsiv-mulga:
## Problem (mulga-siv-265)

Follow-up to #333. E2E Nightly run [27244177680](https://github.com/mulgadc/spinifex/actions/runs/27244177680) cell-18 `TestRebootResilience` failed: after a host reboot, OVN `dnat_and_snat` for **auto-assigned** public IPs and **ELBv2 floating** IPs reconverged to the centralised shape (`external_mac` empty), so all three external IPs (`.231/.232/.233`) went 100% unreachable north-south (`ALB post-reboot: 0/2 healthy`).

#333 fixed this only for store-backed Elastic IPs, by persisting the ENI MAC in `EIPRecord`. The reboot suite allocates no EIP, so that MAC-carrying path is bypassed.

## Root cause

`policy.AddEIP` enables the distributed shape only when `eip.MAC != ""`. For non-EIP public IPs the reconcile spec carried no MAC → `distributed=false` → idempotent-skip on a `LogicalIP+vpc` match → `external_mac` never programmed.

## Fix

The ENI MAC is **already authoritative in OVN** — the ENI port's `port_security` (`apply.go`, `manager_live.go`), per ADR-0006 I1/I2. So:

- `policy.AddEIP` resolves the MAC from `GetLogicalSwitchPort(eip.PortName).PortSecurity[0]` when the caller supplies none — uniformly for EIP, auto-assigned, and ELBv2 IPs. Resolving **before** the idempotency check also forces a stale centralised row to upgrade (delete-then-add) instead of skip.
- `reconcile.loadEIPs` no longer sources the MAC from `EIPRecord`.
- Removed the now-redundant `EIPRecord.MacAddress` field + its set/clear. The associate-time `vpc.add-nat` event still carries the MAC directly (fast path).

## Invariants (ADR-0006)

- S5 — L3/policy never create/delete OVN objects: `GetLogicalSwitchPort` is a read. `network/invariants/policy_no_create_test.go` green.
- I1/I2 — `port_security` MAC is authoritative: this reads that authority instead of duplicating it.
- S3/S9 unchanged.

## Testing

`make preflight` green (lint, vet, race, govulncheck, invariants). New unit `TestNATManager_AddEIP_Distributed_ResolvesMACFromLSP`. E2E: cell-18 reboot suite.

### DIFF
--- a/spinifex/handlers/ec2/eip/service_impl.go
+++ b/spinifex/handlers/ec2/eip/service_impl.go
@@ -213,7 +213,6 @@ func (s *EIPServiceImpl) AssociateAddress(input *ec2.AssociateAddressInput, acco
 	record.InstanceId = instanceID
 	record.PrivateIp = privateIP
 	record.VpcId = vpcID
-	record.MacAddress = macAddr
 	record.State = "associated"
 
 	data, err := json.Marshal(record)
@@ -271,7 +270,6 @@ func (s *EIPServiceImpl) DisassociateAddress(input *ec2.DisassociateAddressInput
 	record.InstanceId = ""
 	record.PrivateIp = ""
 	record.VpcId = ""
-	record.MacAddress = ""
 	record.State = "allocated"
 
 	data, err := json.Marshal(record)
--- a/spinifex/handlers/ec2/eip/types.go
+++ b/spinifex/handlers/ec2/eip/types.go
@@ -9,20 +9,15 @@ const (
 
 // EIPRecord represents a stored Elastic IP address allocation.
 type EIPRecord struct {
-	AllocationId  string `json:"allocation_id"`
-	PublicIp      string `json:"public_ip"`
-	PoolName      string `json:"pool_name"`
-	AssociationId string `json:"association_id,omitempty"`
-	ENIId         string `json:"eni_id,omitempty"`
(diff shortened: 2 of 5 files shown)


joshsiv-mulga:
Closing — live RCA on a single-node mirror host (chassis bottlebrush) + cell-18 artifact analysis proves this fix is the wrong layer.

cell-18 runs CENTRALISED NAT (veth uplink → NATModeFromUplinkMode → centralised; NAT rows have external_mac/logical_port empty, the *correct* shape). policy.AddEIP's distributed external_mac resolve added here NEVER executes in centralised mode, so it is a no-op for this failure. Confirmed: the post-fix nightly run 27245789780 reproduces the identical failure.

True root cause: the reboot suite fails at `WaitForTargetsHealthy: 0/2 healthy` because the ALB system VM's lb-agent heartbeat to the gateway (https://<host>:9999) times out post-reboot. vpcd's only centralised recovery action is a passive `ip neigh flush` (policy primeReachability) — it never proactively re-establishes/verifies north-south reachability, so the guest→gateway path doesn't self-prime within the 5-min window in the heavier CI cell. Re-scoping under mulga-siv-265.

PR #341, opened 2026-06-10
fix(reconcile): demote post-reboot gateway SB claim verify to nudge-once + enrich diagnostic

joshsiv-mulga:
Follow-up to #339.

#339 added an SB Port_Binding chassis-claim verify after `rebindGatewayChassis` with a 30s×N poll + ERROR on non-convergence. CI run `27253185469` showed the recompute **nudge** restores centralised-gateway forwarding (0/2 → 2/2 healthy, both backends served), but the **verify predicate false-negatives** for this gateway shape: no chassisredirect port exists, and `ovn-sbctl find Port_Binding ... --columns=chassis` reads empty even while forwarding works. Result: 90s wasted spin + false ERROR logs on a healthy dataplane.

## Change
- Demote `ensureGatewayClaimed` to **nudge ovn-controller recompute once**, single ~3s recheck, **WARN** (not ERROR) if still unconfirmed — no 30s×N spin.
- Enrich the SB query to dump `logical_port/type/chassis/up/requested_chassis` so the next reboot-suite run reveals the correct claimed-vs-forwarding signal before the predicate is hardened.
- Extract `gatewayClaimArgs` + add `chassisClaimed`/`gatewayClaimArgs` unit coverage.

The recompute nudge — the mechanism proven to restore forwarding — is unchanged. Only the verify/log behaviour is corrected.

ADR-0006: re-asserts existing NB bindings + reads/nudges SB/controller; no NB object create/delete (S5 respected).

Validation: single-node host can't repro (recovers forwarding too fast); next reboot-suite nightly is the read-out via the enriched diagnostic.

### DIFF
--- a/spinifex/network/host/gateway_claim.go
+++ b/spinifex/network/host/gateway_claim.go
@@ -23,24 +23,48 @@ func NewGatewayClaimProber(sbAddr string) *GatewayClaimProber {
 	return &GatewayClaimProber{sbAddr: sbAddr}
 }
 
-// GatewayPortClaimed reports whether the SB Port_Binding for lrpName has a
-// non-empty chassis. An unclaimed binding (chassis == []) means ovn-controller
-// has not installed flows for the gateway redirect, so the floating IPs behind
-// it are unreachable. ctx is part of the verifier contract; ovn-sbctl is a
-// short-lived local query and is not cancelled mid-flight.
-func (p *GatewayClaimProber) GatewayPortClaimed(_ context.Context, lrpName string) (bool, error) {
-	args := []string{"--no-leader-only"}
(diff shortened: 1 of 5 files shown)


joshsiv-mulga:
Closing: RCA of run 27254808154 confirms this approach is ineffective. The post-reboot failure is ovn-controller on the live chassis not re-claiming ANY local Port_Binding (chassis-identity churn), not a missing recompute nudge — `inc-engine/recompute` recomputes flows but does not force port-binding re-claim, and NB `gateway_chassis` is already correct. Superseded by the diagnostic in #342 (merged); root-cause fix tracked under mulga-siv-268/271 pending the next reboot-suite capture.

PR #371, opened 2026-06-16
fix(viperblockd): seal volume block-map to predastore on detach + durability e2es

juliansommer:
## Summary

Split off from `feat/e2e-data-integrity-p0`. Carries the viperblock detach-seal fix together with the two e2e tests that validate it (they fail without this fix on a node that does not hold the volume's local WAL).

## The Bug

Bytes written to an attached EBS volume were lost across a control-plane detach→reattach when the reattach landed on a node without the volume's local WAL (always true multinode; intermittent single-node). The data
chunks reached predastore but the **block→object map** that indexes them did not, so a fresh node hit `mount: bad superblock`.

The only path that seals the block map is `vb.Close()`, and the detach path never guaranteed it fired — it raced `KillProcess` and the `available` transition.

## The Fix

- `ebs.unmount` (viperblockd) now drives a synchronous, waited-upon `vb.Close()` seal to predastore **after** nbdkit exits, mirroring the mount-time VB config (same `BaseDir`, predastore config, master key).
- The seal result is surfaced through `UnmountOne` (now returns `error`) and the unmount NATS timeout is raised to cover the S3 round-trip.
- `DetachVolume` gates the `available` transition on the seal — on failure the volume stays attached/retryable and the local WAL is retained (`Close()` is fail-safe: it only removes local files after a successful predastore write).
- Bulk teardown keeps tolerating per-volume seal errors so terminate stays idempotent

## Performance — detach-seal cost (measured, single-node)

The seal adds a synchronous predastore round-trip to detach, so the `available` transition is now slower. Cost = fixed overhead + work proportional to synced-but-unsealed data at detach time.

| Un-sealed data | Detach (mean of 3) |
| --- | --- |
| 0 MB (floor)   | **4.02 s** |
| 64 MB          | 6.66 s |
| 256 MB         | 13.92 s |
| 512 MB         | 24.31 s |

- Pre-fix baseline (no seal): mean **3.20 s** (bare nbdkit teardown).
- Fixed overhead: **~+0.8 s** (LoadState + LoadBlockState + SaveState +
  SaveBlockState + `RemoveLocalFiles`), independent of data size.
- Data-proportional: **≈0.0395 s/MB (~25 MB/s)**; fit `detach ≈ 4.02 s + 0.0395·MB`.

This is the durability work the missing `Close()` was skipping — replaying + uploading the unsealed chunks — not new overhead. Real-world detach cost scales with how much synced-but-unsealed data a volume holds.


brynmailer-mdc:
> The only path that seals the block map is vb.Close(), and the detach path never guaranteed it fired — it raced KillProcess and the available transition.

Is vb.Close() still sometimes skipped? I'm not fully across this area of the code base. Are we essentially just doubling up on the seal logic, and accepting that the first attempt may fail (depending on a race), but the second attempt will succeed?

brynmailer-mdc:
> > The only path that seals the block map is vb.Close(), and the detach path never guaranteed it fired — it raced KillProcess and the available transition.
> 
> Is vb.Close() still sometimes skipped? I'm not fully across this area of the code base. Are we essentially just doubling up on the seal logic, and accepting that the first attempt may fail (depending on a race), but the second attempt will succeed?

Nevermind, I see now. This PR is essentially just ensuring `vb.Close()` gets called.

spinifex/services/viperblockd/viperblockd.go
```diff
@@ -231,6 +232,55 @@ func openVolumeVB(cfg *Config, volumeName string) (*viperblock.VB, error) {
...
+	if isAuxVolume(volumeName) {
+		return false
+	}
+	_, err := os.Stat(filepath.Join(baseDir, volumeName))
+	return err == nil
+}
+
+// sealVolumeVB persists a detached volume's block->object map to predastore.
+// The runtime nbdkit plugin is the only path that flushes the map on close and
+// it does not reliably fire on detach, so without this seal a reattach on a
```
brynmailer-mdc: Is there a race on the nbdkit plugin side? Seems like proper detaching should be a concern of Viperblock itself no?

juliansommer: yeah probably I just didn't want to mess with viperblock, and issue only affects unmount (which is handled in spinifex viiperblockd.go) so I thought it could fix it spinifex side but I might need to do it in viperblock if this doesn't work

brynmailer-mdc: I'm OK with this for now, if it works. Take a moment to capture the invariant that would be upheld by Viperblock in an ideal world though. Something like "unmount/remount across any two nodes must present a byte-identical view of the volume as it was pre unmount". Not sure on the wording. Stick it in a new section of viperblocks design.md.

Viperblock will be rewritten eventually, and core properties like this are hard to determine until you run into issues like the bug this PR is addressing. Better to capture them so we don't have to run into the problems again post re-write. Plus, they translate extremely well to prop tests.

PR #414, opened 2026-06-24
feat(ecs): ecs-agent skeleton + containerd image pull (Sprint 4c)

joshsiv-mulga:
## Summary

Sprint 4c of the ECS v1 epic (mulga-cs-ecs, GH #81). Stands up the **ecs-agent**
binary that runs inside an ECS container instance (the guest ECS-AMI VM):

- Resolves host identity from **IMDS** (account / instance-id / AZ).
- Connects to the cluster **NATS** bus and **registers** at boot, then
  **heartbeats** every 30s on the Layer-2 subjects (Q14).
- Pulls task images through a **containerd**-backed `ImagePuller` behind an
  interface, with an **ECR resolver** that mints `GetAuthorizationToken`
  credentials from the instance role.

## Shared extraction

The IMDS-credential and ECR-auth flow is lifted out of `ecr-credential-provider`
into `internal/imdscreds` and `internal/ecrauth`, so the agent and the kubelet
provider share one implementation. `ecr-credential-provider` now delegates
(behaviour-identical, guarded by its existing `main_test.go`).

## Bus source of truth

`spinifex/handlers/ecs/bus` holds the Layer-2 subject builders + wire payloads,
shared by the agent now and the scheduler in Sprint 4e.

## Notes

- `build-ecs-agent` is **not** in the default `build`: the agent ships in the
  ECS-AMI guest (sibling bead mulga-siv-432), not the host bundle, and carries the
  heavy containerd dep.
- **genproto ambiguous-import fix:** `go.opencensus.io -> grpc@v1.33.2` drags the
  pre-split monolithic `genproto` into the graph, colliding with the split
  `genproto/googleapis/rpc` module `containerd/ttrpc` needs. A plain `require` is
  dropped by `go mod tidy`, so it's pinned via a **`replace`** to a post-split
  version. The `replace` survives `go mod tidy` — contributors can tidy freely.
- Real containerd Pull from the live ECR path is deferred to live verify inside a
  booted ECS-AMI VM (CI has no containerd).

## Scope boundaries

4c does **not** subscribe `assign.*` or publish `task-state.*` — that lands in
Sprint 4e (scheduler bin-pack + RunTask).

Plan: \`docs/development/archive/feature/ecs-4c-agent-containerd.md\`


joshsiv-mulga:
### Boot failsafes from runtime verify (576177f3)

Ran the agent on the local box (isolated NATS + fake IMDS, since dev NATS is TLS+auth and the agent connects plain). Three rough edges found at the boot surface, now fixed:

- **Lazy ECR gateway client.** `New()` built the gateway HTTP client eagerly, so a missing/malformed `ECS_GATEWAY_CA` killed the agent *before* it could register — even though register/heartbeat never touch the gateway. Now built lazily on first image pull; verified the agent registers + heartbeats with a bogus CA path.
- **Truthful containerd boot check.** `containerd.New()` dials lazily, so a dead socket slipped past the "containerd unavailable at boot" warn and only surfaced on first pull. Added an `IsServing` probe (3s timeout) in `runtime.New`; the warn now fires correctly (`... not serving: ... DeadlineExceeded`).
- **Env-tunable heartbeat.** `ECS_HEARTBEAT_INTERVAL` (default 30s) — verified a 2s override producing register→heartbeat at +2s.

Runtime evidence (bad CA + 2s heartbeat + dead containerd socket):
```
WARN  ecs-agent: containerd unavailable at boot, image pulls disabled err="... not serving: ... DeadlineExceeded"
INFO  ecs-agent: registered cluster=default instance=i-0verify123 subject=ecs.bus.000000000001.default.instance-register.i-0verify123
RECV  ecs.bus.000000000001.default.instance-register.i-0verify123  | {... identity + capacity ...}
RECV  ecs.bus.000000000001.default.instance-heartbeat.i-0verify123 | {"status":"ACTIVE","runningTasks":0, +2s}
```

preflight green, diff coverage 100%.

PR #457, opened 2026-06-30
feat: OVSDB RAFT cluster primitives in setup-ovn.sh (Phase A)

joshsiv-mulga:
Phase A of the OVN OVSDB RAFT clustering epic (mulga-siv-486).

## Summary

Teach `setup-ovn.sh` to bring up the OVN NB/SB databases as a native OVSDB RAFT cluster, so a multinode Spinifex cluster no longer has the init host as a single point of failure for the OVN control plane. This phase adds only the node-level primitive; ansible bootstrap (Phase B) drives it across the cluster.

## Changes (`scripts/setup-ovn.sh`)

- New flags `--db-cluster-local-addr=IP` / `--db-cluster-remote-addr=IP`. Empty remote ⇒ create the cluster (init node); set ⇒ join it.
- When a local addr is given, write a single shared `OVN_CTL_OPTS` to `/etc/default/ovn-central` carrying both NB and SB cluster local/remote addrs (RAFT ports default to NB 6643 / SB 6644), then restart the per-DB units (`ovn-ovsdb-server-nb`, `ovn-ovsdb-server-sb`, `ovn-northd`) directly — the `ovn-central` aggregator is `ExecStart=/bin/true` and won't restart its children.
- `set-connection ptcp:6641` / `ptcp:6642` runs only on the create node; RAFT replicates it cluster-wide, and a joining node would just redirect to the leader.
- Guards: `--db-cluster-local-addr` requires `--management`; `--db-cluster-remote-addr` requires a local addr.
- `--ovn-remote` already passes through verbatim, so a comma-separated SB endpoint list works for compute nodes pointing at the cluster — documented in usage + examples.
- `--help` now prints the full header block (was truncated at line 50).

## Compatibility

The single-node `--management` path is unchanged when no cluster addrs are supplied — dev boxes (`dev-install.sh`) and existing single-node clusters keep working.

## Testing

- `bash -n` syntax check; both arg guards verified to fire.
- `--help` renders the new flags + examples.
- `make preflight` green.
- Full RAFT cluster formation (3 servers, one leader) is exercised end-to-end in Phase B (ansible bootstrap) and Phase E (multinode e2e) — not independently runtime-tested here, since forming a cluster requires the ansible classification + DB `.db` reset that Phase B provides.

### DIFF
--- a/scripts/setup-ovn.sh
+++ b/scripts/setup-ovn.sh
@@ -22,8 +22,16 @@
 #   --mgmt-cidr=CIDR     IPv4 CIDR to assign on the mgmt bridge (default: 10.15.8.1/24)
 #   --mgmt-iface=NAME    Physical/virtual NIC to enslave to the mgmt bridge (multi-node only)
 #   --no-mgmt-bridge     Skip mgmt bridge provisioning (for dev-networking hosts)
-#   --ovn-remote=ADDR    OVN SB DB address (default: tcp:127.0.0.1:6642)
+#   --ovn-remote=ADDR    OVN SB DB address; accepts a comma-separated list of
+#                        SB endpoints for compute nodes pointing at a RAFT
+#                        cluster (default: tcp:127.0.0.1:6642)
 #   --encap-ip=IP        Geneve tunnel endpoint IP (default: auto-detect)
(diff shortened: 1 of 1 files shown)


joshsiv-mulga:
Superseded by #458 — consolidating all epic mulga-siv-486 phases onto a single `feat/ovn-raft` branch.

PR #492, opened 2026-07-09
feat(dns): forwarder hardening, Route53 quota consts, cross-tenant reconcile backstop

benduncan:
Builds on `feat/route53` (Phase B) with three additions, each committed separately.

## 1. Per-tap VPC DNS forwarder hardening (`a9709582`, bead mulga-56q7i)
Closes the resilience regression vs the superseded OVN-LB path:
- **H1** — bind at vpcd startup before the first reconcile (was: rebind only on the reconcile tick, blipping guest DNS on every restart).
- **H2** — per-tap concurrency cap + token-bucket rate limit on UDP/TCP, at AWS resolver-QPS parity, so one flooding/broken guest can't exhaust vpcd.
- **H3** — backend health memory: a failed backend is marked down with a cooldown and tried only after healthy ones, so a dead `backend[0]` stops costing every query a full timeout.

Tests cover the restart window, load-shed, and failover ordering.

## 2. Route53 / Resolver quota consts (`89ce0a6c`)
`handlers/dns/quotas.go` mirrors the AWS service quotas as consts (records-per-zone 10,000, resolver QPS-per-IP 10,000, hosted zones, VPCs-per-zone, change-batch limits, …) behind a `Quotas` struct ready for a future config override. The writer now rejects an UPSERT that would add a **new** record set past the records-per-zone cap (replacing an existing set is never capped).

## 3. Cross-tenant reconcile backstop (`ca9f6ccf`, bead mulga-1005 V1.4)
Periodic loop that rebuilds managed records from the live inventory and converges the zone — repairs a dropped lifecycle event, a writer restart, or an LWW race.

**Multi-tenant aware.** Spinifex hosts multiple tenants that share the base zone, so pruning is gated per record class on a *complete cross-account* enumeration:
- Load balancers (`ListLoadBalancers`, all accounts) and EKS clusters (all `eks-account-*` buckets) are pruned only when their listing succeeds in full.
- EC2 records (node-local `vmMgr` view) are **never** pruned — the terminate hook owns removal.
- A transient store error suppresses that class's pruning rather than delete another tenant's live records ("don't sync one side of the equation").

Changes apply through the same queue-group writer as the lifecycle hooks, so every node running the loop serialises on one writer and never races the zone object — no leader election needed.

### Follow-up
Per-tenant zone *isolation* (accounts currently share the base zone, so DNS is cross-tenant visible) is tracked in bead **mulga-bl5ga** — the reconcile makes the sync tenant-complete; isolation is the open half.

## Testing
`GOWORK=off` (worktree/go.work): build + `golangci-lint` (0 issues) + `go vet` + `go test -race` green for `handlers/dns`, `handlers/elbv2`, `handlers/eks`, `daemon`.


benduncan:
Merged `origin/dev` into this branch (commit `d6793593`), resolving conflicts to preserve both dev's context-threading refactor and the route53 resolvable-DNS work. Verified with `GOWORK=off`: build, vet, golangci-lint, manifest-check, manifest-lint, govulncheck, and the unit/race/harness suites all pass.

PR #563, opened 2026-07-22
feat(config): enable viperblock chunk GC in rendered node configs

joshsiv-mulga:
## Summary

Chunk GC has been off on every deployed node since it was plumbed through. This renders `gc_enabled = true` in the node config template so it is actually on.

## The gap

`GCEnabled` was threaded to every `viperblock.VB` construction site, but nothing ever set it. It is a `*bool` that every consumer treats as off when nil:

| Site | Default when unset |
| --- | --- |
| `spinifex/config/config.go:133` | `GCEnabled *bool` — nil |
| `cmd/spinifex/cmd/service.go:303` | `gcEnabled := false` |
| `handlers/ec2/volume/service_impl.go:207` | nil-guarded false |
| `viperblock/nbd/viperblock.go:79` | `var gc_enabled bool = false` |

The config template renders `shardwal` but omitted `gc_enabled`, despite the two being introduced as siblings. No `.toml` template in the repo contained the key, and there is no occurrence anywhere in `ansible/`. Confirmed on a live node — nbdkit's own command line carried `gc_enabled=false`, and the rendered `[nodes.<host>.viperblock]` block had no such key.

So superseded chunks were never reclaimed under overwrite churn on any deployed node.

## Test

`TestSpinifexTomlTemplate_GCEnabled` renders the template and loads it through `config.LoadConfig`, asserting a non-nil `true`. Matching the template text would not be enough: the value only matters if it parses into a non-nil pointer, since nil is indistinguishable from off at every consumer. Mutation-checked — removing the template line fails the test on the nil assertion.

## Deployment note

The template governs install and bootstrap, so **existing deployments keep GC off until their config is regenerated**. This is not a hot toggle.

**Order matters:** do not deploy GC-on to a node whose viperblock predates the read-modify-write atomicity fix (viperblock #42). GC increases chunk churn and that bug corrupts on the RMW path.

## Known limitation, not addressed here

`ensureGCSnapshotSafe` caches its snapshot-ancestry answer for the life of the VB, and the latch that disables GC when a snapshot appears is in-process only. Spinifex creates snapshots of running volumes from a *different* process (`handlers/ec2/image/service_impl.go`, `snapshotRunningVolume`), so a snapshot taken during a mount does not reach the sweeping engine. Closing that needs a cheap per-volume generation marker checked before each sweep; tracked separately.

### DIFF
--- a/cmd/spinifex/cmd/admin_test.go
+++ b/cmd/spinifex/cmd/admin_test.go
@@ -545,6 +545,32 @@ func TestSpinifexTomlTemplate_EncryptionKeyFile(t *testing.T) {
 	assert.NotContains(t, string(data2), "encryption_key_file", "blank key must omit the field")
 }
 
+// A rendered node config must turn chunk GC on. Asserting the template text
+// alone is not enough: the value has to survive parsing into a non-nil true
+// pointer, because every consumer treats nil as off. Rendering and loading
+// together is the only check that covers the whole path.
+func TestSpinifexTomlTemplate_GCEnabled(t *testing.T) {
+	dir := t.TempDir()
+	settings := admin.ConfigSettings{
+		Node: "node1", Az: "ap-southeast-2a", Port: "4432", Region: "ap-southeast-2",
+		BindIP: "0.0.0.0", AccessKey: "AKIATEST", SecretKey: "SECRET",
+		AccountID: "123456789012", NatsToken: "token",
+		OVNNBAddr: "tcp:127.0.0.1:6641", OVNSBAddr: "tcp:127.0.0.1:6642",
+		ConfigDir: dir,
+	}
+
+	path := filepath.Join(dir, "spinifex.toml")
+	require.NoError(t, admin.GenerateConfigFile(path, spinifexTomlTemplate, settings))
+
+	cfg, err := config.LoadConfig(path)
+	require.NoError(t, err)
+
+	node, ok := cfg.Nodes["node1"]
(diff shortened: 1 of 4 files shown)


joshsiv-mulga:
Hold this until mulga-gowza lands.

This PR renders `gc_enabled = true`, and `GCInterval` defaults to 5 minutes (`viperblock.go:50`, applied at `:1018` whenever the value is zero), so enabling the flag starts a periodic sweep with no operator action.

That makes the cross-process snapshot window live rather than theoretical. `ensureGCSnapshotSafe` caches its snapshot-ancestry answer for the VB lifetime, and the GC-off latch in `snapshot.go:70` is per-process. The engine that sweeps is nbdkit; the engine that would trip the latch is viperblockd. Spinifex snapshots running volumes from the daemon (`handlers/ec2/image/service_impl.go`, `snapshotRunningVolume`), so the latch never reaches the sweeper and the sweeper can delete chunks the new snapshot pins.

CI reaches this directly: `tests/e2e/single/createimage_test.go` builds a custom AMI from a *running* instance, which is exactly the reproduction. Already-deployed nodes are unaffected by a merge (the template only applies to newly rendered configs), but every CI e2e env bootstraps fresh, so the first run after merge would render the flag on.

Failure mode is a silently corrupt AMI surfacing as an intermittent boot failure on a 5-minute race, which is expensive to attribute. mulga-jgz3n is now marked as blocked by mulga-gowza to keep this ordering visible.

PR #566, opened 2026-07-22
test: complete the e2e live-vs-mock rebalance (epic mulga-8oyg3)

joshsiv-mulga:
Completes epic `mulga-8oyg3`, 11/11. The integration tier gates the live tier, and a control-plane-only diff now finishes CI without provisioning a tofu env.

## What is here

**A durability test that can actually fail.** `tests/e2e/partialblock` replaces the `memdurability` suite, which passed twice on a build that still lost data.

The engine only enters its read-modify-write path for a *partial* block. The old workload wrote through ext4 with 1 MiB buffers, which the filesystem coalesced into aligned full-block requests — block-layer tracing recorded zero unaligned and zero sub-4096 requests, so the path under test never ran, and writers 4 MiB apart could not collide on a block even in principle. It also could not have failed if it had reached it: its write phase ended in `wait; sync; sha256sum` and reported `sha256sum`s exit status, while bare `wait` returns 0 unconditionally, so writer errors were invisible and a backend ENOSPC presented as a pass.

The replacement drives two concurrent `O_DIRECT` writers into disjoint halves of the same 4096-byte blocks on a raw unformatted device, stamps every half with a round number, then drops the page cache and rereads with `O_DIRECT` asserting every half carries the final round. Each writer runs as its own process with a real exit status, and both the status and the exact byte count are asserted before any durability claim is made.

**Fail-first, measured on env12** — same node, same test binary, same workload, engine the only variable:

| engine | result |
| --- | --- |
| `viperblock v1.13.0` (pre-fix) | FAIL — `CHECKED 8192 BAD 1758` (21.5%) |
| `viperblock f8cfd77` (#42) | PASS — `CHECKED 8192 BAD 0`, 74.4s |

Every mismatch carried the predicted signature: `got_round=5 want_round=6`, one half of a block a generation behind while its partner held the current round. The verifier was also mutation-checked before either live run — injecting a stale stamp made it exit 1 naming the exact block, half and round.

On that evidence the suite is in the **default** suite list, not dispatch-only. Cost per run is 8 MiB of region and ~0.1G of backend, against the predecessor 2048 MiB working set and roughly 3 GiB leaked per run that twice drove a node under the nearfull watermark.

**ECR data plane moved into the front gate.** `tests/integration/ecr_registry_test.go` covers push/pull round trip with byte-identical blob and manifest comparison, refusal to auto-create a repository on push, and retag without disturbing the original tag. Three tests, 0.18s, no guest boot.

This closes `mulga-7m84e` by invalidating its premise rather than implementing it. That bead ruled out the in-process gate because the tests "bind fixed ports via a real predastore"; the gate as built serves from an `httptest.Server` on an ephemeral port, so concurrent runs on one host cannot collide. The capacity/load leg deliberately stays live — it exists to put real parallel pressure on a real predastore, which a memory-backed object store cannot stand in for.

**Chunk GC stays wired on** — `CreateVolume` now asserts it, closing `mulga-u5nx1`.

**Backend byte accounting descoped** to node-wide growth in `storagegrowth`. A per-volume footprint needs an S3 listing that returns 403 on every volume: predastore enforces a same-account bucket-ownership check on top of IAM, and the harness authenticates as a tenant-scoped identity against a system-account-owned bucket. Granting the harness a policy-check-skipping credential was rejected as inverting the trust boundary for test convenience. Revisit under `mulga-89l1s`.

## Correction to the history in this branch

`8ea5a3d7` reverts the memdurability suite, and its commit message gives the wrong reason — it says the test compared a warm-cache read against its own write. That is not accurate: the suite did run `drop_caches` between the two hashes. It deserved removal for the reasons above (swallowed writer errors, and a whole-block workload that never reaches the RMW path), recorded in `mulga-i10ts`. Left unamended rather than rewriting pushed history.

## Not in scope

- No final before/after `e2e_single` wall-clock. Only the first-18-tests measurement exists (8.9s of 1989s, 0.45%), consistent with the plan doc prediction that the payoff is a 35s front gate rather than reclaimed budget.
- Coverage equivalence accepted on parity with what the old **default** suite list ran, that being the coverage which actually gated merges.

## Verification

`make preflight` green. Live runs on env12 as tabled above; new ECR tests mutation-checked. Branch carries the fixed viperblock pin, so the newly-gating suite runs against an engine it passes on.

Closes mulga-8oyg3, mulga-7m84e, mulga-i10ts, mulga-u2353, mulga-u5nx1, mulga-iwqkb.


joshsiv-mulga:
Pushed `22806be0` — fixes `mulga-17238` (E2E volume teardown leaks volumes).

The fixture registered `DeleteVolume` as a bare, unconditional call, so a suite creating its instance before its volume had the volume torn down first, while still attached: `VolumeInUse`, nothing retried, volume stranded. Reordering the LIFO chain was rejected — registration order cannot express the instance/volume dependency. Teardown is now self-sufficient for any creation order: force-detach, wait for `available`, then delete. Instance teardown waits for `terminated` for the same reason.

Two things kept the leak silent and are also fixed: a failing cleanup was logged and forgotten (now reported as a leak that fails the test, or surfaces through `Close() error` for TestMain to fail the run), and a panic in one cleanup aborted the whole chain (each callback is now isolated).

Verification: 5 new unit tests in `tests/e2e/harness`, mutation-checked — restoring the bare `DeleteVolume` reproduces the exact reported failure, and removing the panic isolation crashes the run. `make preflight` passed. Live on env12: partialblock PASS in 73.8s, fixture volume deleted cleanly, instance reached `terminated`, node left with zero volumes and zero live instances.

PR #591, opened 2026-07-28
feat: match new predastore config surface area

brynmailer-mdc:


cmd/spinifex/cmd/templates/predastore-multinode.toml
```diff
@@ -3,8 +3,8 @@
 version = "4"
 region = "{{.Region}}"
 
-host = "0.0.0.0"
-port = 8443
+# The gateway listen address comes from the daemon CLI options: a top-level
```
juliansommer, line 6: probably need migrations for this

PR #626, opened 2026-08-01
fix(vm): report ENOSPC to the guest instead of pausing the VM, and let the stuck-terminate backstop fire

joshsiv-mulga:
Two contained storage-exhaustion defects, both reproduced live on env13 on 2026-08-01. Both bead descriptions named the wrong mechanism; the corrected root causes are recorded on the beads and in the plan doc.

Plan: `docs/development/bugs/enospc-report-and-stuck-terminate-stamp.md`

## mulga-xwsm7 — a data-volume ENOSPC pauses the whole VM

QEMU's default `werror=enospc` pauses the VM on an out-of-space write. The boot drive already avoids this with `Werror: "report"`; data volumes never got the same treatment, in either the cold-boot or the hotplug path.

Worth stating explicitly, since it is a behaviour change: with `report`, a guest filesystem on the affected volume typically goes read-only rather than the VM pausing. That is the AWS-like behaviour and matches what the boot drive has always done.

What the repro refuted, since it sent the original RCA down the wrong path:

- QMP is **not** wedged. It answers in 0.0s and reports `status: io-error`, with `io-status=nospace` on the affected drive only.
- Nothing hangs. The ENOSPC mapping is synchronous with no retry loop, and the VM resumes cleanly once space is freed.

## mulga-t4hz1 — the stuck-terminate backstop is disabled by the condition it guards

On a full filesystem the local state write fails, so `transitionWithPrecheck` returns before `stampShuttingDownAt` runs — while the in-memory status has already moved to `shutting-down`. `StuckTerminateReaper.Sweep` skips any VM whose `ShuttingDownAt` is zero, so the 10-minute backstop never fires. Waited past the deadline live; the force-complete line never appeared.

The consequences are not cosmetic: the wedge is permanent (freeing 25 GiB did not recover it), it leaks instance capacity, and after a daemon restart a requested terminate had silently become a `stopped` instance with all three volumes still attached.

This stamps on the persistence-failure path as well. The error is still returned to the caller, and the raced `ErrInvalidTransition` branch deliberately does **not** stamp, since the target state was never entered there.

This does not conflict with the proposed `BeginStopOrTerminate` split in `docs/development/bugs/stop-terminate-ack-before-state-transition.md` — that routes through the same helper and relies on the stamp reaching the backstop.

## Testing

`make preflight` passes. New tests, each verified to fail before the change:

- data volumes carry `werror=report,rerror=report` cold-boot and hotplug, boot drive unchanged
- a failed state write still stamps `ShuttingDownAt` and still returns the error
- the raced branch does not stamp
- the reaper force-completes a VM stamped via the failure path — the case currently skipped forever

Closes mulga-xwsm7, mulga-t4hz1.

### DIFF
--- a/spinifex/vm/devices.go
+++ b/spinifex/vm/devices.go
@@ -105,6 +105,11 @@ func HotplugEBSBus(port int) string {
 // AttachVolume's QMP device_add (rendered as a map) and buildDrives' -device
 // (rendered as a joined command-line string) so the two block-graph shapes
 // cannot diverge.
+//
+// werror/rerror mirror the boot drive's on-error policy (see Drive.Werror):
+// a data volume has no -drive-backed BlockBackend to carry it, so the qdev
+// device itself sets werror=report,rerror=report to report backend ENOSPC to
+// the guest instead of letting QEMU's default werror=enospc pause the VM.
 func VolumeBlkDeviceArgs(volumeID, nodeName, iothreadID, bus string) []string {
 	return []string{
 		fmt.Sprintf("id=%s", VolumeDeviceID(volumeID)),
@@ -112,6 +117,8 @@ func VolumeBlkDeviceArgs(volumeID, nodeName, iothreadID, bus string) []string {
 		fmt.Sprintf("iothread=%s", iothreadID),
 		fmt.Sprintf("serial=%s", VolumeSerial(volumeID)),
 		fmt.Sprintf("bus=%s", bus),
+		"werror=report",
+		"rerror=report",
 	}
 }
 
@@ -123,7 +130,8 @@ func VolumeBlkDevice(volumeID, nodeName, iothreadID, bus string) Device {
 }
 
 // VolumeBlkDeviceQMPArgs renders the same virtio-blk-pci arguments as a
-// device_add argument map for AttachVolume's QMP hot-attach path.
+// device_add argument map for AttachVolume's QMP hot-attach path. See
+// VolumeBlkDeviceArgs for why werror/rerror are set here.
 func VolumeBlkDeviceQMPArgs(volumeID, nodeName, iothreadID, bus string) map[string]any {
 	return map[string]any{
 		"driver":   "virtio-blk-pci",
@@ -132,6 +140,8 @@ func VolumeBlkDeviceQMPArgs(volumeID, nodeName, iothreadID, bus string) map[stri
 		"iothread": iothreadID,
 		"serial":   VolumeSerial(volumeID),
 		"bus":      bus,
+		"werror":   "report",
+		"rerror":   "report",
 	}
 }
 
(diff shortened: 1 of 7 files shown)


joshsiv-mulga:
## Live verification on env13 — both fixes PASS

Deployed this branch to env13 (clean `cluster-reset` first) and ran both repros against the deployed binary.

### mulga-xwsm7 — PASS

Confirmed the running qemu qtree actually carried `werror=report,rerror=report` on the data-volume device before drawing any conclusion. Then ballasted the store to ~5% free (the old 15% nearfull band no longer fires) and drove `dd if=/dev/urandom` against the data volume, polling every 2s for ~3.5 minutes:

- guest CPU jiffies **kept climbing** throughout — the pre-fix signature was jiffies dropping to ~0 and staying there
- `query-status` stayed `{"running": true, "status": "running"}` for the whole run, with **no `STOP` event**
- journal showed `BLOCK_IO_ERROR` x3, on the data volume only, and **not** followed by `STOP`
- inside the guest: `dd` failed with `No space left on device`, that filesystem went read-only, and the guest stayed fully responsive on SSH and other I/O

Boot drive behaviour unchanged.

### mulga-t4hz1 — PASS

Filled the filesystem holding `/var/lib/spinifex` to 0 bytes free and terminated a running instance. The local state write failed with ENOSPC exactly as expected — that is the trigger, not a problem — but `ShuttingDownAt` was stamped anyway, so the instance stayed visible to the reaper.

Waited the full ~10 minute `stuckTerminateTimeout`, polling and timestamping throughout rather than concluding early. At the deadline:

```
vm/gc: force-completing terminate wedged past timeout
```

and the instance reached `terminated` — not `stopped`, which is where it landed pre-fix.

### One interaction worth recording

The JetStream ENOSPC latch (mulga-e9u4o, tracked separately) did activate and blocked *persistence* of the terminated record, with repeated `Failed to write terminated instance to KV` / `context deadline exceeded`. It did **not** affect the reaper's detection or action: `kvHealthy()` never tripped, so GC sweeps kept running throughout. New detail for that bead — freeing disk space alone did not clear the latch even 15+ minutes later; only a `spinifex-nats` + `spinifex-daemon` restart did.

### Two unrelated defects found while verifying

Both filed rather than folded into this PR, since neither is in this diff's theme:

- **mulga-26avl** — data volumes stay `in-use` after their instance reaches `terminated`, and become unrecoverable: `detach-volume --force` fails `InvalidInstanceID.NotFound` because the instance record is already gone, and `delete-volume` fails `VolumeInUse`. Reproduced twice, via both a normal terminate and a reaper-triggered force-complete. This is the durable form of the "volume never frees" symptom mulga-t4hz1 described — the terminate now completes correctly, but the volumes still do not release.
- **mulga-gyoml** — an orphaned QEMU process survived both a `cluster-reset` and a daemon restart, absent from `describe-instances` the whole time.

env13 was reset afterwards and is clean: no instances, no volumes, all 13 units active, 40% disk.

PR #643, opened 2026-08-01
fix: stop lbagent heartbeats on positive LB deletion, and actually enable the JetStream watchdog

joshsiv-mulga:
Closes **mulga-adph7**. Advances **mulga-e9u4o**, which stays open pending a live fill-and-recover
test (see below).

Two components that latch into a bad state and never recover on their own.

## 1. lbagent heartbeats a deleted load balancer forever (`mulga-adph7`)

The agent retried indefinitely after its LB was deleted, filling the log with noise after teardown.

**The risk in fixing this is over-correcting.** Stopping on any heartbeat error would let a transient
NATS blip or a control-plane restart permanently kill a healthy agent — much worse than log noise. So
the agent stops only on a positive, unambiguous deletion signal:

`LoadBalancerNotFound` (`awserrors.ErrorELBv2LoadBalancerNotFound`), parsed from the gateway's
`<ErrorResponse><Error><Code>` envelope. `handlers/elbv2/service_impl.go:1043` returns that code **only**
when `store.GetLoadBalancer` round-trips successfully and returns `nil` — a positive "no such record"
answer. A store *error* (NATS/KV unreachable) returns `ErrorServerInternal` instead, so a dial
failure, timeout or 5xx can never be mistaken for deletion.

There is a second ambiguity worth calling out: a **fresh** LB can also see `LoadBalancerNotFound`
before its record replicates to the gateway node fielding the first heartbeat. So deletion is only
inferred once the existing `registrationGrace` window has passed, or a heartbeat has already succeeded
once. Before that it stays a startup race and the agent keeps retrying.

`tick()` also now no-ops if `stopCh` is already closed, closing a Start-loop race that could emit one
extra heartbeat after stopping.

### Tests

- `TestHeartbeat_NotFoundWithinGraceDoesNotStop` / `..._NotFoundAfterGraceStopsAgent` — the grace
  carve-out, both directions.
- `TestHeartbeat_NotFoundAfterSuccessStopsAgent` — the primary scenario: heartbeating fine, then the
  LB is deleted.
- `TestHeartbeat_TransientErrorDuringOperationDoesNotStop` — repeated 503s after a successful
  heartbeat must **not** stop the agent. This is the one guarding against the over-correction.

Two pre-existing tests had incidentally used `LoadBalancerNotFound` as their generic-failure fixture;
they are repointed at a real 500, since ambiguous failure is now a distinct code path from deletion.

## 2. The JetStream ENOSPC watchdog was never switched on (`mulga-e9u4o`)

**Where the latch lives:** entirely inside vendored nats-server, not spinifex.
`nats-server/v2@v2.14.3/server/jetstream.go:619` `handleOutOfSpace` calls `go s.ShutdownJetStream()`
on a no-space write error, setting an internal `disabled` flag (`jetstream.go:664`) that only a
process restart clears. Its own doc comment says it is "for transient runtime errors that the
operator is expected to fix before restarting". `monitor.go:3648` `healthz` returns 503 on
`?js-enabled-only=true` once set.

So clearing the latch in-process is not ours to do — there is no public re-enable API and it would
mean patching vendor code.

**This is where it differs from `mulga-mv3mt`**, the viperblock `backendFull` latch, which looks like
the same bug: viperblock owns that flag as its own Go state and can clear it on a recovery probe. That
approach was deliberately *not* mirrored here, because the state is not ours.

**The detect-and-restart path already existed and was dead code.**
`spinifex-nats-watchdog.timer`/`.service` and `build/scripts/nats-js-watchdog.sh` landed in
`da33e574` — the script samples `healthz` three times over ~6 s and only then `try-restart`s
`spinifex-nats`. But `scripts/setup.sh` `install_systemd()` drops the unit file and enables
`spinifex.target` **without ever enabling the timer**, and `WantedBy=timers.target` only self-activates
once enabled. `cmd/installer/firstboot/firstboot.go` had the identical gap on the ISO path.

So the watchdog has never run on any real deployment — which is exactly why "a restart is the only way
out" was the observed behaviour. Both paths now `systemctl enable --now spinifex-nats-watchdog.timer`.

Anti-flap logic is untouched: 3 samples with 2 s gaps, `try-restart` only when the service is already
`is-active` so it never fights an in-flight start/stop, 60 s boot delay, 30 s repeat.

### Still pending

The live fill-and-recover test is **not** done — no environment was available. It matters more than
usual here: the watchdog script has never run in anger, so the live test is what would catch it being
*broken* as well as disabled.

1. Deploy; confirm `systemctl is-enabled spinifex-nats-watchdog.timer` → `enabled`, and it appears in
   `systemctl list-timers`.
2. Fill the JetStream `StoreDir` volume to 0 free.
3. Confirm `journalctl -u spinifex-nats` logs "JetStream out of ... resources, will be DISABLED" and
   `curl 127.0.0.1:8222/healthz?js-enabled-only=true` returns 503.
4. Free the space.
5. Watch `journalctl -u spinifex-nats-watchdog` for "unhealthy on 3 consecutive probes ...; restarting",
   and confirm KV writes resume unaided within one timer cycle (~30-90 s).

## Preflight

Passes: manifest-check/lint OK, golangci-lint 0 issues, govulncheck clean, test-cover and
test-harness pass.

Note the diff-coverage step reported "No new/modified Go source lines" because it ran against
`origin/dev` before these commits existed locally; CI will diff the pushed commits properly.

### DIFF
--- a/cmd/installer/firstboot/firstboot.go
+++ b/cmd/installer/firstboot/firstboot.go
@@ -197,6 +197,9 @@ fi
 
 # Enable services to start, on reboot
 systemctl enable spinifex.target spinifex-banner.service
+# WantedBy=timers.target only self-activates once enabled — without this the
+# JetStream ENOSPC-latch watchdog never runs and a full disk needs a manual restart.
+systemctl enable --now spinifex-nats-watchdog.timer
 
 # Mark complete only after every step above has succeeded. Until this point,
 # any failure (set -e) leaves the marker absent and the next reboot retries
--- a/cmd/installer/firstboot/firstboot_test.go
+++ b/cmd/installer/firstboot/firstboot_test.go
@@ -114,6 +114,25 @@ func readScript(t *testing.T, root string) string {
 	return string(b)
 }
 
+// TestWriteScriptEnablesNatsWatchdogTimer pins that firstboot activates the
+// JetStream ENOSPC-latch watchdog timer. The unit file's WantedBy=timers.target
+// only takes effect once enabled — without this line the watchdog is dropped
+// onto disk but never runs, and a full disk needs a manual nats restart forever.
+func TestWriteScriptEnablesNatsWatchdogTimer(t *testing.T) {
+	root := t.TempDir()
+	makeRootDirs(t, root)
+
+	cfg := Config{Hostname: "test-node", ClusterRole: "init"}
+	if err := Write(root, cfg); err != nil {
+		t.Fatalf("Write: %v", err)
+	}
+	content := readScript(t, root)
+
+	if !strings.Contains(content, "systemctl enable --now spinifex-nats-watchdog.timer") {
+		t.Error("firstboot script must enable --now spinifex-nats-watchdog.timer")
+	}
+}
+
 func TestWriteScriptCallbackAfterDoneMarker(t *testing.T) {
 	root := t.TempDir()
 	makeRootDirs(t, root)
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -663,6 +663,10 @@ EOF
     fi
     $SUDO systemctl daemon-reload
     $SUDO systemctl enable spinifex.target
+    # The unit file alone does not self-activate (WantedBy=timers.target only
+    # takes effect once enabled) — without this the JetStream ENOSPC-latch
+    # watchdog is inert and a full disk requires a manual restart forever.
+    $SUDO systemctl enable --now spinifex-nats-watchdog.timer
     info "Systemd units installed and enabled (per-service users)"
 }
 
--- a/spinifex/lbagent/agent.go
+++ b/spinifex/lbagent/agent.go
@@ -10,6 +10,7 @@ import (
 	"crypto/tls"
 	"encoding/hex"
 	"encoding/xml"
+	"errors"
 	"fmt"
 	"io"
 	"log/slog"
@@ -25,6 +26,7 @@ import (
 
 	"github.com/mulgadc/spinifex/internal/gwsign"
 	"github.com/mulgadc/spinifex/internal/tlsconfig"
+	"github.com/mulgadc/spinifex/spinifex/awserrors"
 )
 
 const (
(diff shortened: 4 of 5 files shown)


joshsiv-mulga:
Correction to this PR's stated root cause for `mulga-e9u4o`, after a live fill-and-recover test on env13 against this merged code (`34aa6d22`).

**The latch this PR describes is not the one that fires.** This PR attributed the ENOSPC latch to `handleOutOfSpace` → `ShutdownJetStream()` → the global `disabled` flag, detectable via `healthz?js-enabled-only=true` returning 503. The live test never produced that: `healthz` stayed `200 {"status":"ok"}` throughout, and no `will be DISABLED` line was ever logged.

**What actually fires is the per-`msgBlock` `werr` latch** in vendored `server/filestore.go`. `werr` is assigned at 4 sites (`:5108`, `:7232`, `:8293`, `:8318`) and cleared at **none** — `grep 'werr = nil'` returns zero hits in the file. `writeMsgRecordLocked:7214` returns it immediately on every later write ("Return previous write errors immediately"), so one ENOSPC write wedges that stream permanently until the process restarts, regardless of disk recovery.

`healthz?js-enabled-only=true` reads the *global* JetStream flag, which `werr` never sets. **So the watchdog cannot observe this failure mode by construction** — it is not mistuned, it is probing the wrong signal.

Live evidence (env13): filled `/dev/vda1` to 0 with a 30.24 GiB ballast, reproduced the exact original signature — `Filestore [KV_spinifex-cluster-state] Critical write error: ... no space left on device` — then freed the space and watched unaided for 4m20s across 9 watchdog cycles. The identical ENOSPC text kept being emitted with 29G genuinely free, `healthz` stayed 200, and `ExecMainStartTimestamp` never changed: no restart was ever attempted.

**What in this PR still stands:** enabling `spinifex-nats-watchdog.timer` was correct and necessary. It was dead code on every deployment ever made, and it is now confirmed enabled and running cleanly every ~30-35s with all exits 0. That part is unchanged and worth keeping.

**What is still needed:** the probe must attempt a real JetStream *write* (a KV put/get round-trip on a dedicated canary key), since a write is the only operation that observes `werr`. That is a design decision rather than a script tweak — the node ships no `nats` CLI and `spx` exposes no KV probe subcommand — so it is tracked on `mulga-e9u4o`, which remains open.

The `mulga-adph7` half of this PR (lbagent heartbeat stop on positive LB deletion) is unaffected.

PR #649, opened 2026-08-02
fix(gateway): log the three SigV4 paths that reject silently

joshsiv-mulga:
Advances **mulga-zjy2g** (does not close it — see below).

## What the evidence actually shows

`mulga-zjy2g` has survived three sweeps of static tracing through `sigv4` on the theory that ImportKeyPair's public-key body triggers a canonicalisation or body-rewind defect. Two independent signer lineages (aws-sdk-go-v2 and v1) have now been fuzzed against the canonicalisation and both pass. This PR comes from evidence instead.

**Live A/B on env13** — three OpenTofu configs reproducing the s3-webapp provider shape exactly (explicit static creds, `s3` endpoint pointed at predastore, `s3_use_path_style`, and the real apply ordering with the S3 bucket and IAM role created before the key pair). All three applied cleanly, as did a direct `aws ec2 import-key-pair`. The request shape does not reproduce.

**The CI artefact** gives the exact failure and, critically, a gateway-minted RequestID:

```
Error: importing EC2 Key Pair (s3-webapp-demo): operation error EC2: ImportKeyPair,
https response error StatusCode: 403, RequestID: c8c83f87-9077-49c2-b73d-3dd49aa50f0f,
api error SignatureDoesNotMatch
```

`writeSigV4Error` mints that RequestID via `uuid.NewString()`, so the 403 came from this middleware. Three call sites emit `SignatureDoesNotMatch`:

| `auth.go` | Path | Logged before this PR |
|---|---|---|
| `:57` | `ErrRequestTimeTooSkewed` | **no** |
| `:70` | service not in `supportedServices` | **no** |
| `:121` | `sig.Verify` mismatch | yes (Warn) |

Elasticsearch has the failing run (`ci_run_id 30467638631`, 1.45M docs, retention covers it). awsgw shipped 141 INFO and 17 DEBUG records for it and **zero WARN/ERROR** — and DEBUG from the same service, host and run did ship, so the pipeline is not filtering by level. A `:121` rejection would have produced a WARN. It did not. The failure therefore came through one of the two silent paths, and nothing on disk can say which.

That is why the bead's own documented next step — "diff the logged canonical request against the one the SDK signed" — could never execute: it assumed the failure arrived via the one path that logs.

## Change

Both silent paths now log at Warn. The skew path records the client timestamp, the server timestamp and `MaxClockSkew`, which is what distinguishes clock drift from a signing defect. The unsupported-service path names the rejected service.

`signingTime` is needed because `Parse` rejects a skewed request before returning a `SignedRequest`, so the raw header (or presigned query arg) is the only copy of the timestamp left to log.

## Tests

Three tests in `auth_silentmismatch_test.go`, all mutation-checked — silencing either log line fails the corresponding test with `"" does not contain ...`, i.e. the empty output that is exactly the pre-fix behaviour.

`make preflight` passes.

## Why the bead stays open

Criterion 1 ("ImportKeyPair through the terraform provider is accepted") is unproven — the failure is intermittent and did not reproduce on demand. This PR makes the next occurrence diagnosable in one shot rather than another sweep of guesswork.

### DIFF
--- a/spinifex/gateway/auth.go
+++ b/spinifex/gateway/auth.go
@@ -64,7 +64,10 @@ func (gw *GatewayConfig) SigV4AuthMiddleware() func(http.Handler) http.Handler {
 					gw.writeSigV4Error(w, r, awserrors.ErrorSignatureDoesNotMatch)
 				default:
 					// Malformed Authorization, bad credential scope, unsupported
-					// algorithm, missing content hash: a failed auth attempt.
+					// algorithm, missing content hash: a failed auth attempt. The parse
+					// error names which, and is the only record of it.
+					slog.Warn("Auth failure: malformed signature envelope",
+						"sourceIP", clientIP, "err", err)
 					gw.RateLimiter.RecordFailure(clientIP)
 					gw.writeSigV4Error(w, r, awserrors.ErrorIncompleteSignature)
 				}
--- a/spinifex/gateway/auth.go
+++ b/spinifex/gateway/auth.go
@@ -52,7 +52,14 @@ func (gw *GatewayConfig) SigV4AuthMiddleware() func(http.Handler) http.Handler {
 				case errors.Is(err, sigv4.ErrPayloadTooLarge):
 					gw.writeSigV4Error(w, r, awserrors.ErrorRequestEntityTooLarge)
 				case errors.Is(err, sigv4.ErrRequestTimeTooSkewed):
-					// Skew/replay: AWS returns this as SignatureDoesNotMatch.
+					// Skew/replay: AWS returns this as SignatureDoesNotMatch, which is
+					// indistinguishable on the wire from a canonicalisation mismatch.
+					// Log the timestamps so the two can be told apart after the fact.
+					slog.Warn("Auth failure: request time too skewed",
+						"sourceIP", clientIP,
+						"requestTime", signingTime(r),
+						"serverTime", time.Now().UTC().Format("20060102T150405Z"),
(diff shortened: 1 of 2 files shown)


joshsiv-mulga:
Extended to a third path after checking against **mulga-hhrn5** (the existing gateway-observability feature bead).

`auth.go`'s parse-error `default:` branch answers `IncompleteSignature` for a malformed Authorization header, unsupported algorithm or missing content hash — and was likewise silent. That made it the last auth rejection path with no record of which validation stage rejected the request. It now logs the parse error, which names the stage.

So all three previously-silent paths are covered, and every auth rejection in this middleware now leaves a log line.

Also worth recording, since it corrects a hypothesis on hhrn5: that bead assumed the gap was the shipping pipeline dropping WARNs. It isn't — awsgw shipped 141 INFO **and 17 DEBUG** docs for run 30467638631 from the same service and host, so the pipeline passes every level. The lines simply were never emitted.

And hhrn5's motivating incident (an ECR-cache `403` from a us-east-1/ap-southeast-2 region mismatch) would in fact have logged: `parseCredential` only structurally validates the credential scope and never compares `Region` to the gateway's, so a region mismatch reaches `sig.Verify` — the one branch that already logged. Its invisibility in ES traces to hhrn5's now-closed dependency `mulga-ju6bf`, where the OTel pipeline dropped slog attributes.

New test is mutation-checked like the others: silencing the line fails with `"" does not contain "malformed signature envelope"`. `make preflight` passes.

PR #653, opened 2026-08-02
fix: close the storage-exhaustion recovery gaps (durable terminate, per-service memory caps, local volume leak)

joshsiv-mulga:
Closes mulga-t4hz1 (P1, residual), mulga-wupqz (P2), mulga-anvjn (P2).

Theme: a node whose storage has filled cannot fully recover itself, and nothing bounds what any one service consumes. All three were confirmed by live verification on env12 before any code was written.

> **Pending:** a `go.mod` pin bump to the merged viperblock revision will be pushed to this branch before merge. Please hold until it lands.

## 1. Terminate finalises in memory but never durably — mulga-t4hz1 residual

PR #626 fixed the headline deadlock: under ENOSPC, `transitionWithPrecheck` stamps `ShuttingDownAt` even when `WriteState` fails, and the stuck-terminate reaper force-completes 10 minutes later. Re-verified live and still working.

What remained: `finalizeTerminated` bailed at `transitionWithPrecheck(instance, StateTerminated)` before reaching `StateStore.WriteTerminatedInstance`, so the durable terminated record was never written. In-memory status still flipped to `terminated`, which dropped the instance into a gap where nothing owned it — `StuckTerminateReaper` skipped it (no longer `shutting-down`) and `TerminatedTeardownReaper` never saw it (it only scans `ListTerminatedInstances()`). Confirmed live: 5+ minutes after freeing the disk, zero retry activity. Only a daemon restart recovered it.

`finalizeTerminated` now distinguishes `ErrInvalidTransition` from a persistence-only failure. On the latter it continues into the durable KV write, `DeleteIf`, hooks and `writeRunningState`, then returns the original transition error. Chosen over a new "in-memory terminated, not durable" reaper because it needs no new state or scan surface and the existing reaper is already idempotent.

Live proof of the branch being taken, with the root filesystem at genuinely 0 bytes free:

```
17:27:55 WARN "Local state persistence failed on terminate, continuing to durable KV write" instanceId=i-30c7d241e64deb85b err="...no space left on device"
```

And unattended recovery on a second instance — no restart, no operator action:

```
17:19:35 ERROR Failed to delete volume ... InsufficientStorage
17:21:35 INFO  DeleteVolume completed / Deleted volume on termination
```

**Known remaining hole, deliberately not fixed here:** if the JetStream KV write also fails — which happens when NATS's file store shares the exhausted disk — the early return still orphans the instance. That needs control-plane state isolation and is recorded against mulga-e9w3w.

## 2. No per-service memory cap — mulga-wupqz

`spinifex.slice` sets `MemoryMax=90%` / `MemoryHigh=80%`, but every service under it was uncapped. Measured live: predastore at 1.2G idle, all four services `MemoryMax=infinity`. One runaway exhausts the slice budget and the slice-level cap then reaps an arbitrary victim rather than the offender.

Added percentage-based `MemoryHigh`/`MemoryMax` to all 9 long-running units — daemon 75/85 (guest QEMU still shares its cgroup), predastore 40/55, viperblock 15/25, the rest 10/20. Percentages rather than byte counts so the same units work on a 15G dev box and on large bare metal. Verified live via `systemctl show -p MemoryHigh -p MemoryMax`.

The per-service maxima intentionally sum past 100%. cgroup v2 enforces the minimum at every level, so the slice's 90% stays the real binding ceiling; the per-service caps add value by catching a leak in one of the smaller services before it grows enough to trip a slice-wide OOM and take an unrelated victim with it.

**Correction to the bead's own wording:** it says "loopback/NBD kernel cache", which is wrong for the production path. Guest volumes attach as `-drive file=nbd:unix:...sock`, QEMU's userspace NBD client. There is no `/dev/nbdX` in the production I/O path — that appears only in dev and import scripts. This PR is scoped to the cap gap only.

## 3. DeleteVolume leaves the local volume directory — mulga-anvjn

Root-caused to spinifex, not viperblock. The `ebs.delete` handler stopped the volume's goroutines, killed nbdkit and removed the socket, but never removed the local WAL/checkpoint cache. The only path that did — `VB.Close`'s `RemoveLocalFiles` — runs from the unmount seal step, which `-efi` volumes never reach and which a main volume also skips after a failed seal. Measured ~96MB retained per instance, paid twice because the `-efi` companion leaks identically. `/var/lib/spinifex` is not a separate mount on a stock deploy, so this accrues against host root.

Verified live: after terminate, both directories are gone, while a separately attached `DeleteOnTermination=false` volume correctly keeps its directory.

## 4. Path validation, from review

The first version of the cleanup passed `filepath.Join(cfg.BaseDir, ebsRequest.Volume)` straight to `os.RemoveAll` with the volume name unmarshalled from the NATS message and never validated. An empty name collapses the join to `BaseDir` and wipes every volume's local state on the node; `filepath.Join` cleans `..` rather than rejecting it, so a traversal escapes `BaseDir` entirely. Far worse than the leak being fixed.

Added `localVolumeDir(baseDir, volume)`, which rejects empty, `.`, `..`, any name containing a separator (`filepath.Base(name) == name`) and an empty `baseDir`, then independently re-verifies containment with `filepath.Rel`. Applied as a gate in `ebs.delete`, `ebs.mount` (which ran `clearStaleSealReceipt` on the unvalidated name) and `ebs.config` (whose not-mounted branch derives the same path). Each rejects before touching the filesystem and returns `Success: false` with an error rather than skipping silently.

`ebs.unmount` was checked and is not independently at risk — it only acts on `matched.Name` from `cfg.MountedVolumes`, which is populated solely by the now-gated mount handler. `volumeNeedsSeal` was routed through the same helper as defence in depth.

## Tests

- `TestTerminate_LocalPersistenceFailure_StillDurablyRecordsTerminated` — A/B proven; reverted, `store.terminated[v.ID]` is nil.
- `TestIntegration_EBSDeleteRemovesLocalVolumeDirectory` — A/B proven.
- `TestIntegration_EBSDeleteEmptyVolumeNameDoesNotWipeBaseDir` — A/B proven; unpatched, the log confirms `os.RemoveAll` targeted `BaseDir` itself.
- `TestIntegration_EBSDeleteRejectsPathTraversal`, `TestIntegration_EBSDeleteValidNameLeavesSiblingsUntouched`, `TestValidVolumeName`, `TestLocalVolumeDir`.

The `../..` case was deliberately **not** executed against the unpatched handler: under `TMPDIR=/var/tmp` the traversal resolves to `/var/tmp` itself, so a real A/B run would have wiped a shared directory. The empty-name case proves the same `os.RemoveAll(joinedPath)` code path destructively, and the traversal arithmetic is covered non-destructively by the unit test.

No Go test fits the systemd units; those were verified live instead.

`make preflight` clean: `golangci-lint: 0 issues`, `govulncheck: no vulnerabilities`, all tests green.

Plan doc: `docs/development/bugs/storage-exhaustion-recovery-gaps.md`.

### DIFF
--- a/build/systemd/spinifex-awsgw.service
+++ b/build/systemd/spinifex-awsgw.service
@@ -19,6 +19,10 @@ RestartSec=5
 TimeoutStopSec=10
 
 OOMScoreAdjust=-500
+# Per-service ceiling bounds a leak in this service alone rather than
+# tripping the slice-wide cap, which would pick an arbitrary victim.
+MemoryHigh=10%
+MemoryMax=20%
 Environment=SPINIFEX_CONFIG_PATH=/etc/spinifex/spinifex.toml
 Environment=SPINIFEX_BASE_DIR=/var/lib/spinifex/awsgw/
 Environment=SPINIFEX_AWSGW_TLS_CERT=/etc/spinifex/server.pem
--- a/build/systemd/spinifex-daemon.service
+++ b/build/systemd/spinifex-daemon.service
@@ -19,6 +19,10 @@ KillMode=process
 TimeoutStopSec=15
 # RG-4: infra tier — reaped only after every guest
 OOMScoreAdjust=-500
+# Guest QEMU still runs inside this cgroup, so the ceiling tracks the slice
+# budget (just under 80%/90%), not just daemon overhead.
+MemoryHigh=75%
+MemoryMax=85%
 Environment=SPINIFEX_CONFIG_PATH=/etc/spinifex/spinifex.toml
 Environment=SPINIFEX_BASE_DIR=/var/lib/spinifex/spinifex/
 Environment=SPINIFEX_WAL_DIR=/var/lib/spinifex/spinifex/
--- a/build/systemd/spinifex-nats.service
+++ b/build/systemd/spinifex-nats.service
@@ -17,6 +17,10 @@ RestartSec=5
 # stops before nats does, so size this for nats's own graceful-stop needs.
 TimeoutStopSec=20
 OOMScoreAdjust=-500
+# Per-service ceiling bounds a leak in this service alone rather than
+# tripping the slice-wide cap, which would pick an arbitrary victim.
+MemoryHigh=10%
+MemoryMax=20%
 Environment=SPINIFEX_CONFIG_PATH=/etc/spinifex/nats/nats.conf
 
 # Filesystem access
--- a/build/systemd/spinifex-northstar.service
+++ b/build/systemd/spinifex-northstar.service
@@ -17,6 +17,10 @@ RestartSec=5
 # NATS unsubscribe/close and process exit that follow it.
 TimeoutStopSec=10
 OOMScoreAdjust=-500
+# Per-service ceiling bounds a leak in this service alone rather than
+# tripping the slice-wide cap, which would pick an arbitrary victim.
+MemoryHigh=10%
+MemoryMax=20%
 Environment=SPINIFEX_CONFIG_PATH=/etc/spinifex/spinifex.toml
 Environment=SPINIFEX_BASE_DIR=/var/lib/spinifex/northstar/
 
--- a/build/systemd/spinifex-predastore.service
+++ b/build/systemd/spinifex-predastore.service
@@ -22,6 +22,11 @@ RestartSec=5
 # backstop if that ordering is ever bypassed (direct stop, future edit).
 TimeoutStopSec=30
 OOMScoreAdjust=-500
+# Legitimately the most memory-hungry infra service (S3 buffering/caching);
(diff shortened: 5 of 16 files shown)


joshsiv-mulga:
Pin bump landed in `d7a5c57e`. `spinifex/go.mod` now resolves to `github.com/mulgadc/viperblock v1.14.1-0.20260802084514-ed52f6a0e4c0`, confirmed with `GOWORK=off go list -m` to be the merge commit of viperblock#68 (`ed52f6a`). `GOWORK=off go build ./...` and a full `make preflight` both pass on the pinned branch — lint 0 issues, govulncheck 0 reachable, diff coverage 65.4% against a 60% threshold. Ready to merge.

PR #663, opened 2026-08-02
feat(installer): three-plane networking, ZFS root, distributed NAT and multi-node install

benduncan:
## Summary

Consolidates the ISO installer, multi-node networking, ZFS root and distributed
NAT work into one branch for landing on `dev`. Supersedes `feat/iso-installer`,
`feat/multi-node-networking`, `feat/public-ip-distributed-nat` and
`feat/zfs-iso`, all of which are contained here.

The through-line is making a node installable and clusterable without hand
editing: the installer now understands that a box has more than one NIC and
that those NICs have distinct roles, and everything downstream — listeners,
resolvers, OVN, NAT — follows from that role assignment rather than from a
single `--bind` guess.

## Changes

**Three-plane networking**

- wan/lan/vpc network roles with VLAN and MTU, plus a role-assignment TUI with a
  hardware NIC table and an Advanced toggle.
- Public listeners and resolvers stay on the wan plane even when `--bind` names a
  specific address; the S3 API stays on every plane in that case.
- A DHCP wan address is resolved at first boot so `--advertise` has something real.
- NICs are identified from sysfs rather than `networkctl`; plane bridges inherit
  their port's MAC; bridge ports no longer gate `network-online`.

**ZFS root**

- ZFS root filesystem with per-disk bootloaders, and a swap file sized from RAM.
- The daemon unit picks up `host.env`, which raises `SPINIFEX_RESERVED_MEM_GB` by
  the ARC ceiling the admission gate cannot otherwise see.

**Distributed NAT**

- Distributed NAT on veth uplinks; EIPs advertise from the gateway port rather
  than the localnet; no more probing a link-local gateway LRP.

**Clustering**

- `install-node.sh` plus a VPC-building smoke test that spreads instances across
  nodes via a placement group; `--wipe` and a standalone `node-reset.sh`.
- Cluster formation retries the join while the primary is still starting, refuses
  to join with an already-initialized node unless forced, runs joins concurrently
  and clears renamed chassis rows.
- OVN RAFT bring-up no longer silently serves a standalone database; the chassis
  name is pinned to the cluster node name; `setup-ovn.sh` refuses to clobber a
  non-OVS link holding the mgmt bridge name.
- The ISO's Join Existing Cluster option is removed — it could not work as built.

## Merge notes

Merged `origin/dev` (46 commits). Three conflicts, each where both sides touched
adjacent lines:

- `scripts/setup.sh` — kept dev's `spinifex-ui` data directory alongside this
  branch's idempotent awsgw config symlink. Both additive.
- `firstboot_test.go` — dev's new watchdog-timer test set `ClusterRole`, removed
  here along with the Join Existing Cluster option. Dropped the field; `"init"`
  was the default path the test already exercised.
- `spinifex-daemon.service` — both sides added a two-line comment to a file
  already at the RG-11 twelve-comment budget. Condensed both to one line rather
  than dropping either rationale.

## Testing

`make preflight` passes: golangci-lint 0 issues, govulncheck clean, manifest
check and drift lint, diff coverage 75.5% against the 60% threshold. `make
test-actions` and `make test-race` (0 data races) also pass.

`make test-integration` fails locally on `TestCreateVolume_PersistsToRealPredastore`,
but this is pre-existing and not from this branch — pristine `origin/dev` fails
identically in a clean worktree, and dev's CI is green. The shared predastore
fixture sets `SSL_CERT_FILE` via `t.Setenv`, which unwinds when the first test
using the fixture ends while the fixture itself stays shared; it only reproduces
locally because Go caches the system cert pool process-wide. Worth its own issue.

Touches `spinifex/network/`, so this needs a CODEOWNERS review. The ADR-0006
invariant tests under `spinifex/network/invariants/` are unmodified and pass.


cmd/installer/install/install.go
```diff
@@ -62,90 +78,83 @@ func Run(cfg *Config) error {
...
+		{"format ESPs", func() error { return formatESPs(cfg.Storage) }},
+		{"load zfs module", loadZFSModule},
+		{"create zfs pool", func() error {
+			opts = resolveZFSOpts(cfg.Storage)
+			slog.Info("zfs pool", "topology", cfg.Storage.FS, "disks", len(cfg.Storage.Disks),
+				"ashift", opts.Ashift, "compress", opts.Compress, "arc_max_mib", opts.ARCMaxMiB)
+			return createPool(cfg.Storage, opts)
+		}},
+		{"create zfs datasets", createDatasets},
+		{"configure zfs", func() error { return writeZFSSystemConfig(opts) }},
```
joshsiv-mulga, line 112: **`"configure zfs"` runs before `"copy rootfs"`, so `rsync --delete` removes both files it just wrote**

`writeZFSSystemConfig` writes `/etc/modprobe.d/zfs.conf` and `/etc/spinifex/host.env` into the target, but this step sits inside `zfsRootSteps`, which is appended *before* `"copy rootfs"` (line 63). That step is `rsync -aHAX --delete / → mountRoot/`, and neither path is in its exclude list.

The chain, traced statically:

- `writeZFSSystemConfig` has exactly one caller — this line. Nothing re-runs it after the copy.
- The exclude list in `copyRootfs` covers neither `/etc/spinifex` nor `/etc/modprobe.d`.
- Nothing creates either file in the live ISO rootfs. The only `/etc/spinifex` and `/etc/modprobe.d` references under `scripts/iso-builder/` are in `spx-test-gpu.sh`, which reads those paths on an already-installed system.
- Because the rsync source has no `/etc/spinifex` at all, `--delete` removes the directory this step populated, not merely the file. `installSpinifex` later re-creates `/etc/spinifex` for `node.conf`, which is why that file survives and `host.env` does not.

If that holds, the ARC ceiling is never applied and `SPINIFEX_RESERVED_MEM_GB` is never raised: the node runs with the OpenZFS default ARC (half of RAM) and a 2GB reserve.

**Why CI and the install test stay green either way**

`TestWriteZFSSystemConfigDeclaresTheARCCeilingToBothConsumers` calls `writeZFSSystemConfig` directly against a temp root, so it never exercises the step ordering.

`scripts/iso-builder/test/test-zfs.sh` asserts only `installation complete`, `login:`, and `root=ZFS=rpool/ROOT/spinifex`. A node missing both files passes all three — it installs, it boots, and it genuinely is running on the pool. The symptom would surface much later as guests admitted into memory the ARC already holds, which does not look like an install fault.

**Verification — 5 seconds on any node installed from this branch with a ZFS root**

```
ls -l /etc/spinifex/host.env /etc/modprobe.d/zfs.conf
cat /sys/module/zfs/parameters/zfs_arc_max
```

Missing files, with `zfs_arc_max` at roughly half of RAM rather than the capped value, confirms it. Both files present and a capped value means I have this wrong, and I would like to see that machine.

To be clear about what this is: a static trace, not something I have observed on a running node — I do not have a ZFS host to try it on.

**Suggested fix** — either move `"configure zfs"` to after `"copy rootfs"`, or add `--exclude=/etc/spinifex` and `--exclude=/etc/modprobe.d/zfs.conf` in `copyRootfs`. Moving the step looks cleaner, since `opts` is already resolved by that point.

joshsiv-mulga (approved):

PR #678, opened 2026-08-03
fix(imds): answer ARP for the captured link-local addresses

joshsiv-mulga:
## Problem

The per-tap IMDS datapath on `br-imds` installs an ingress demux flow matching `ip,nw_dst=169.254.169.254`, resting on the assumption stated in its own comment:

> The guest addresses the frame to its gateway MAC (.254/.253 are off-link, routed via the default gateway)

That holds for a guest which routes to the address. It does not hold for a guest which treats `169.254.0.0/16` as on-link per RFC 3927 — such a guest ARPs for `169.254.169.254` first. Nothing answered that ARP:

- `br-imds` had no ARP flows, so the request fell through the demux (matching `ip`) to the forward flow and was bridged to `br-int`
- `br-int` had no flow matching `arp_tpa=169.254.169.254`
- OVN has no `localport`, despite a comment in `dhcp_options.go` claiming one answers this

Real EC2 answers this ARP, which is why Windows guests work there.

## Evidence

Packet capture on a Windows Server 2022 guest's tap, after a reboot:

```
16:53:48.388700 02:21:87:a4:05:0c > ff:ff:ff:ff:ff:ff, ARP, Request who-has 169.254.169.254 tell 172.31.0.7
16:53:49.089458 02:21:87:a4:05:0c > ff:ff:ff:ff:ff:ff, ARP, Request who-has 169.254.169.254 tell 172.31.0.7
...
```

39 requests, zero replies. The guest's `cloudbase-init.log` shows the consequence — `ConnectTimeout` to `169.254.169.254`, then `No metadata service found` — so no plugin past metadata discovery runs. Bridge flows were intact and `vpcd` logged `IMDS: tap responder serving` throughout, so this is not a lifecycle fault.

A Linux guest on the same subnet reaches IMDS on every boot, including across a self-initiated reboot, because it routes via the gateway and hits the existing demux flow. That is why the gap went unnoticed.

## Change

Install a per-tap ARP responder for each captured address, under the same cookie as the existing flows so teardown already removes it. The reply advertises `GatewayMAC` — the same L2 identity the egress flow presents — so the guest sees one MAC for the address whichever way it resolved it, and the ingress demux (matching `in_port` and `nw_dst`, not dst MAC) still catches what it sends.

Also corrects the `dhcp_options.go` comment asserting a localport that does not exist.

## Testing

- Unit tests assert a responder per captured address with the right `arp_tpa`, `arp_op`, reply MAC and hex-encoded fields, and that malformed MAC/address input errors rather than emitting a broken flow
- `make preflight` passes

Live validation on env12 is still outstanding and needs a deploy.

Bead: `mulga-8h3gy`

### DIFF
--- a/spinifex/network/host/imds_datapath.go
+++ b/spinifex/network/host/imds_datapath.go
@@ -2,10 +2,12 @@ package host
 
 import (
 	"context"
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"hash/fnv"
 	"log/slog"
+	"net"
 
 	"github.com/mulgadc/spinifex/spinifex/utils"
 )
@@ -22,10 +24,11 @@ const (
 // Kept in one place so the endpoint addresses and the ingress flows stay in sync.
 var imdsCaptureAddrs = []string{imdsMetaAddr, imdsDNSAddr}
 
-// Per-tap OpenFlow priorities on IMDSBridge. Demux (.254/.253 interception) and
-// egress sit above the forward flows so IMDS traffic is captured and everything
-// else is bridged tap<->patch to br-int.
+// Per-tap OpenFlow priorities on IMDSBridge. The ARP responder, demux
+// (.254/.253 interception) and egress sit above the forward flows so IMDS
+// traffic is captured and everything else is bridged tap<->patch to br-int.
 const (
+	imdsARPPriority     = 250
 	imdsDemuxPriority   = 200
 	imdsForwardPriority = 100
 )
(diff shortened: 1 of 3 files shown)


joshsiv-mulga:
Validated live on env12 (`i-93390093153d17b13`), with the flow installed by hand on `br-imds` ahead of a deploy:

```
17:17:24.778171 Request who-has 169.254.169.254 tell 172.31.0.7
17:17:24.778709 Reply 169.254.169.254 is-at 02:f8:82:f2:f4:d4
17:17:24.779073 172.31.0.7.49672 > 169.254.169.254.80: Flags [SEW]
17:17:24.782138 GET /2009-04-04/meta-data/local-hostname HTTP/1.1
```

One request, one reply, then the connection the guest had been unable to open — against 39 unanswered requests before the flow.

The guest then fetched `public-keys/0/openssh-key`, meaning `LocalScriptsPlugin` ran `password-emit.ps1` for the first time on Spinifex, and `aws ec2 get-password-data` returned a 344-char blob that decrypted with the launch key pair.

Still outstanding: deploy this branch so the flow is installed by the code at launch rather than by hand.

PR #681, opened 2026-08-03
fix(ec2): default Windows AMIs to HttpTokens=optional at launch

joshsiv-mulga:
## Problem

`buildMetadataOptions` stamps `HttpTokens=required` whenever a launch omits `MetadataOptions`, regardless of the image. cloudbase-init — the only Windows guest agent we ship — has no IMDSv2 token support in any release: its `EC2Service` pins `_metadata_version = '2009-04-04'` and never issues a `PUT /latest/api/token`. Spinifex's IMDSv1 gate answers an untokened GET only when the instance carries `HttpTokens=optional`, so a Windows instance launched the ordinary way 401s on every metadata request, never leaves `PRE_METADATA_DISCOVERY`, and never runs `LocalScriptsPlugin` — which is what emits the encrypted administrator password.

Every successful Windows launch so far has needed `--metadata-options "HttpTokens=optional"` passed by hand, which is not something a Terraform user or the console flow would know to do.

## Change

The platform signal already exists: `utils.PlatformFromDetails(amiMeta.PlatformDetails)` returns `"windows"` or nil, and `PrepareRunInstances` already stamps it onto `ec2Instance.Platform` right after `RunInstance` returns. `RunInstance` itself resolves no AMI metadata, so the default is applied at the point where the platform becomes known.

- `defaultHTTPTokensForPlatform` returns `optional` for Windows, `required` for everything else.
- `applyPlatformTokenDefault` rewrites the stamped value only when the request said nothing about `httpTokens`, so an explicit value always wins — including a Windows launch that deliberately asks for `required`.
- `validateMetadataOptions` and `applyMetadataOptions` are untouched: `ModifyInstanceMetadataOptions` keeps taking the caller at their word, and the legacy nil-block path has no image in hand.

The documented default stays `required`; Windows is the single platform-derived exception, mirroring how AWS Windows AMIs behave in practice.

## Testing

`make preflight` passes. New unit coverage:

- `TestDefaultHTTPTokensForPlatform` — `optional` for `"windows"`, `required` for nil, empty and an unexpected value.
- `TestApplyPlatformTokenDefault` — flips an unspecified Windows launch, leaves Linux alone, and is a no-op on either platform when the caller named a value.
- `TestApplyPlatformTokenDefaultNilBlock` — a legacy instance with no metadata-options block is left alone rather than panicking.
- `TestPrepareRunInstances_PlatformTokenDefault` — end to end through the launch path over `Windows`, `Windows BYOL` and `Linux/UNIX` images. Confirmed load-bearing: reverting the call site fails the two Windows cases with `expected: "optional" / actual: "required"`.

Plan: `docs/development/bugs/windows-imds-http-tokens-default.md`. Closes `mulga-2dz0d`.

### DIFF
--- a/spinifex/handlers/ec2/instance/metadata_options.go
+++ b/spinifex/handlers/ec2/instance/metadata_options.go
@@ -6,6 +6,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/mulgadc/spinifex/spinifex/awserrors"
+	"github.com/mulgadc/spinifex/spinifex/utils"
 )
 
 // defaultMetadataHopLimit matches the AWS default applied when none is requested.
@@ -14,7 +15,8 @@ const defaultMetadataHopLimit = int64(1)
 // buildMetadataOptions returns the metadata-options block stamped at launch.
 // hopLimit applies only in the AWS-valid 1-64 range, otherwise falling back to
 // the default. An empty httpTokens stamps "required", so an instance that does
-// not ask for IMDSv1 never silently gets it.
+// not ask for IMDSv1 never silently gets it; applyPlatformTokenDefault later
+// relaxes that for the one platform that cannot speak IMDSv2.
 func buildMetadataOptions(hopLimit *int64, httpTokens string) *ec2.InstanceMetadataOptionsResponse {
 	limit := defaultMetadataHopLimit
 	if hopLimit != nil && *hopLimit >= 1 && *hopLimit <= 64 {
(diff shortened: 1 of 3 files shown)


joshsiv-mulga:
Verified live on env12 after `make cluster-deploy`:

| Launch | Platform | HttpTokens |
| --- | --- | --- |
| Windows AMI, no `--metadata-options` | `windows` | `optional` |
| Linux AMI, no `--metadata-options` | none | `required` |
| Windows AMI, explicit `HttpTokens=required` | `windows` | `required` |

The hands-off Windows launch booted through to `LocalScriptsPlugin`, emitted the encrypted blob on the console, and `get-password-data --priv-launch-key` decrypted it. No `--metadata-options` was passed at any point. Test instances terminated.

PR #754, opened 2026-08-09
fix: keep the RDS master password out of the server and console logs

juliansommer:
## Problem

The RDS master password is applied with `ALTER ROLE ... PASSWORD '<cleartext>'`. psql interpolates the literal before the server sees it, so PostgreSQL logs the statement as it received it. Three parameters make that happen:

- `log_statement = ddl` or `all` — modifiable from a customer DB parameter group.
- `log_min_duration_statement = 0` — also modifiable, and logs every completed statement with its text.
- `log_min_error_statement` — not exposed, but its built-in default of `error` logs the statement on *any* failure of the `ALTER ROLE`, so the leak needs no customer parameter at all.

The customer's parameters are already live at that moment: `install_runtime_config` installs `10-rds-parameters.conf` before `bootstrap_master` runs, and the bootstrap postmaster overrides `hba_file`, `listen_addresses`, `unix_socket_*` and `ssl` but not `config_file`. The resulting `postmaster.log` lives on the data volume, so the password propagates into every snapshot and restore. Rotations through `postgresEngine.SetPassword` hit the same logging, on a cluster serving the customer's parameters outright.

A second path is the guest console. `rds-init.initd` sends rds-init's stdout and stderr to `/dev/console`, which `setup.sh` binds to ttyS0 for host-side capture. `psql_bootstrap` runs with `ON_ERROR_STOP=1` and unfiltered stderr, and psql echoes the statement it failed on — so a server error on the `ALTER ROLE` puts the password into the host serial log.

The Go half already redacted the password out of the *error string* it returns, which addresses neither the server's own log nor rds-init's console output.

## Changes

**Session-local logging guard.** Both password-carrying scripts now prepend `SET log_statement = 'none'; SET log_min_duration_statement = -1; SET log_min_error_statement = 'panic';` ahead of the `ALTER ROLE`. All three are SUSET and both sessions connect as the cluster superuser over the socket under peer auth, so the SET succeeds and, being session-local, cannot be overridden by the parameter group. In rds-init the guard sits in `psql_bootstrap` rather than in each caller's heredoc, so no bootstrap statement added later can run without it.

`log_min_duration_statement` is beyond the bead's literal acceptance criteria, but it is the same hole with the same fix.

**Redacted console output.** rds-init routes psql's stdout and stderr through a shell filter that replaces the master password with `[REDACTED]` before it reaches `/dev/console`. The filter uses parameter expansion, not `sed`/`awk`, so the secret never becomes an argv any guest process can read from `/proc`. psql's exit status crosses the added pipeline through a status file in `${BOOTSTRAP_DIR}`, since POSIX `sh` has no `pipefail` — a missing status file reads as a failure rather than a silent success.

## Testing

`make preflight` passes. `make test-images` passes (`shellcheck` is not installed on this host; both changed scripts were checked clean via the `koalaman/shellcheck:stable` container with `-S warning -s sh`).

New cases in `rds-init_test.sh`:

- **log-guard** — bootstraps an instance whose parameter group sets `log_statement = all`, asserts the customer's parameter file really was installed, that all three guard statements reached psql, that *every* bootstrap session carries the guard, and that the guard precedes the `ALTER ROLE`.
- **console-redaction** — the psql stub now echoes the failing statement and its password on stderr the way a real psql does. Asserts the captured rds-init output carries `[REDACTED]` and not the password, that the diagnostic itself survives, and that the failure still propagates (datadir cleared, non-zero exit).

Both were confirmed to fail against the pre-fix script (7 failing assertions) and to pass after. The suite was also run under `dash` and `busybox sh` — the guest runs Alpine's ash, and the redaction relies on quoted patterns inside parameter expansion.

New case in `engine_test.go`: `TestPostgresEngine_SetPasswordSilencesStatementLogging` asserts `SetPassword` sends all three guard statements ahead of the `ALTER ROLE`.

### DIFF
--- a/cmd/rds-agent/engine.go
+++ b/cmd/rds-agent/engine.go
@@ -187,7 +187,13 @@ func (e *postgresEngine) SetPassword(ctx context.Context, username, password str
 	if err := handlers_rds.EnginePostgres().ValidateUsernameNotReserved(username); err != nil {
 		return fmt.Errorf("refusing to set the password of a role the engine reserves: %w", err)
 	}
-	const sql = `\getenv master RDS_MASTER_USERNAME
+	// psql interpolates the password into the ALTER ROLE before the server sees
+	// it, and these three are what would write it to the log — the last on any
+	// failure at its own default. SUSET, so the parameter group cannot win.
+	const sql = `SET log_statement = 'none';
+SET log_min_duration_statement = -1;
+SET log_min_error_statement = 'panic';
+\getenv master RDS_MASTER_USERNAME
 \getenv password RDS_MASTER_PASSWORD
 ALTER ROLE :"master" WITH LOGIN PASSWORD :'password';
 `
--- a/cmd/rds-agent/engine_test.go
+++ b/cmd/rds-agent/engine_test.go
@@ -185,6 +185,40 @@ func TestPostgresEngine_SetPasswordKeepsTheSecretOutOfArgv(t *testing.T) {
 	}
 }
 
+// psql interpolates the password into the ALTER ROLE before the server sees it,
+// so a parameter group that turns statement logging on would write the rotated
+// password into a postmaster.log every snapshot then carries.
+func TestPostgresEngine_SetPasswordSilencesStatementLogging(t *testing.T) {
+	runner := &recordingRunner{}
+	engine := newTestEngine(t, runner.run)
+
+	if err := engine.SetPassword(context.Background(), "mulgamaster", "n3w-pw"); err != nil {
+		t.Fatalf("SetPassword: %v", err)
+	}
+	if len(runner.calls) != 1 {
+		t.Fatalf("ran %d commands, want 1", len(runner.calls))
+	}
+	sql := runner.calls[0].Stdin
+	alter := strings.Index(sql, "ALTER ROLE")
+	if alter < 0 {
+		t.Fatalf("SQL %q does not alter the role", sql)
+	}
+	for _, guard := range []string{
+		"SET log_statement = 'none';",
+		"SET log_min_duration_statement = -1;",
+		"SET log_min_error_statement = 'panic';",
+	} {
+		at := strings.Index(sql, guard)
+		if at < 0 {
+			t.Errorf("SQL %q is missing %q", sql, guard)
+			continue
+		}
+		if at > alter {
+			t.Errorf("SQL %q sets %q only after the ALTER ROLE has already been logged", sql, guard)
+		}
+	}
+}
+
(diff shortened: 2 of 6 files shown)


juliansommer:
Follow-up in ffb5c29e2, from a review of the new code's failure paths. Three real defects in the first commit, all confirmed by reverting each fix in turn:

**The status file was read back unvalidated, and `return ""` is fatal in the guest's shell.** `printf > file` creates before it writes, so a torn write — a full or read-only `/run`, a signal between the two syscalls — leaves the file present and empty. `[ -f ]` passed, the empty value overwrote the deliberate default of 1, and `return ""` ran. busybox ash and dash treat that as a fatal error (`return: Illegal number:`, exit 2), where bash merely returns 2 and carries on — so this would never have reproduced on a dev box. rds-init died inside `psql_bootstrap`: `stop_bootstrap_server` never ran and the `local all all trust` pg_hba was never removed, while the EXIT trap still swept the datadir out from under the still-running postmaster. Reverting the fix and running the suite under `busybox sh` reproduces it exactly — `FAIL: psql-status-empty: bootstrap server left holding the datadir`.

The value is now checked before it is returned, and the missing-file and unreadable-file branches each say so in their own words. Blaming psql for a status file that could not be written would send an operator to a PostgreSQL that did exactly what it was asked.

**`redact_stream` failed open on a multi-line password.** It is line-oriented, so a secret spanning a line boundary matches neither half and both halves print in the clear — the exact case the redaction exists for. This was reachable: `ValidateMasterUserPassword` was bounds-only plus a four-character denylist, so a newline was accepted, and `shellQuote` preserves it across the handoff. Fixed at the boundary — the validator now enforces the printable-ASCII range AWS documents — and rds-init refuses a multi-line password before initdb spends it, so the guest asserts the property rather than silently depending on another service for it.

**The harness wrote the handoff password unquoted**, unlike rds-agent, so no test could have carried a metacharacter through to the checks that exist for it. Now single-quoted the way `shellQuote` writes it.

Also noted and left as-is: the guard shifts psql's `<stdin>` line numbers by three, now documented at the call site; and `log_min_error_statement = 'panic'` applies to `create_database` too, which carries no secret. Narrowing it would reintroduce the risk that a statement added later escapes the guard, so it stays global.

New cases: `psql-status-{empty,garbage,absent}` (status file corrupted through a `cat` stub and an unwritable path), and `multiline-password`. `TestValidateMasterUserPassword` gains the control characters and the non-ASCII range. `make preflight` passes; the suite passes under `bash`, `dash` and `busybox sh`; both scripts are shellcheck-clean via the container.

PR #769, opened 2026-08-12
feat(security): add an nftables host firewall scoping cluster ports to peers

joshsiv-mulga:
Closes the host-firewall half of `mulga-adgld` (proposal 3 of `docs/development/bugs/port-security.md`). Plan: `docs/development/bugs/port-security-charon-and-firewall.md`.

## Problem

Nodes ship no host firewall, so every cluster listener — OVN NB/SB, the formation port, NATS, predastore — is reachable from anything that can route to the node. `install-node.sh` already logged "Restrict those ports with a firewall" and left it at that. On a single-NIC node the planes collapse onto the public address, so that is the default rather than the exception.

## What this does

One `table inet spinifex_filter`, input hook only, `policy drop`. Public ports (22, 3000, 8443, 9999, 53) stay open; cluster ports are scoped to the resolved plane addresses of cluster members, and Geneve/IKE/NAT-T/ESP to the encap plane.

`setup.sh` installs the policy and an apply helper but does **not** enable them. The peer set is unknown until the cluster's planes resolve, and a drop policy with an empty set would cut the first node that applied it out of the formation it is part of. `install-node.sh` writes `/etc/spinifex/firewall/peers.nft` and enables the unit after formation, and preflights every host for `nft` and the helper so a node that cannot take the policy is caught before any host is touched. `--no-firewall` opts out.

## Constraints this respects

- **Never flush the ruleset.** vpcd writes MASQUERADE and per-EIP FORWARD rules into the `ip nat`/`ip filter` tables and reinstalls them only on start. Debian's `nftables.service` flushes on both start and stop, which is why this ships its own unit with no `ExecStop`. The policy create-then-deletes only its own table, in one transaction.
- **INPUT only.** nftables runs every table on a hook and any drop is final, so a forward-hook drop here could not be rescued by vpcd's ACCEPTs in the other table. OUTPUT untouched — the IMDS reply path egresses under per-tap policy routing.
- **`iifname "ime-*" accept` is load-bearing.** IMDS and VPC DNS terminate on per-ENI OVS internal ports in the host netns. Without it, cloud-init, instance-role credentials and all guest DNS break.
- **No `vhp-*` rule, deliberately.** Traced: every dial across a Bedrock host port is daemon-to-guest, so `ct state established` covers the replies. An `iifname "vhp-*" accept` would expose every host service to serving VMs to fix a direction nothing sends in.
- **No guest dials NATS.** lb-agent uses the mgmt gateway URL; the EKS and ECS agents were migrated onto the gateway; rds-agent is explicit that it never uses NATS. That is what makes peer-scoping 4222 safe.

## Verified on env13

Full `make cluster-reset` from this branch, clean bootstrap, `failed=0`.

- Firewall active, cluster healthy, smoke test passes — instance launched, EIP assigned, cloud-init and SSH verified. That exercises the `ime-*` rule, guest DHCP, and the EIP path that proves vpcd's rules still work alongside the new table.
- From off-node, OVN SB 6642 and formation 4432 are blocked while 22 and 9999 still answer. The drop log independently caught another host on the LAN reaching for 6642 and 4432.
- Applying the policy twice in a netns leaves a stand-in `ip filter` FORWARD rule untouched.
- Fail-open confirmed: the first run failed on a missing `nft` and the node stayed reachable and healthy.

Not covered — env13 is single-node, so peer-scoped Geneve/IKE/ESP between chassis and cross-node 5300 forwarding still need a multi-node env.

## Companion mulga PR

`nftables` has to be added in three package paths; the ansible bootstrap play, the gold image and the ISO list live in mulga and land separately. Without that companion change this branch installs the policy but nothing enables it on an ansible-formed cluster.

### DIFF
--- a/build/systemd/spinifex-daemon.service
+++ b/build/systemd/spinifex-daemon.service
@@ -41,9 +41,9 @@ DeviceAllow=/dev/net/tun rw
 DeviceAllow=char-vfio rw
 DeviceAllow=/dev/vfio/vfio rw
 ReadOnlyPaths=/etc/spinifex/spinifex.toml /etc/spinifex/server.pem /etc/spinifex/server.key /etc/spinifex/ca.pem
-# /proc/sys/net/ipv4/conf: per-tap IMDS endpoints set rp_filter/accept_local, which
-# ProtectKernelTunables=yes would otherwise keep read-only.
-ReadWritePaths=/var/lib/spinifex/spinifex /var/log/spinifex /run/spinifex /run/spinifex/nbd /proc/sys/net/ipv4/conf
+# /proc/sys/net/ipv4/conf: ProtectKernelTunables would keep the per-tap IMDS
+# rp_filter/accept_local read-only; /etc/spinifex/firewall: the root helper writes peers.nft.
+ReadWritePaths=/var/lib/spinifex/spinifex /var/log/spinifex /run/spinifex /run/spinifex/nbd /proc/sys/net/ipv4/conf -/etc/spinifex/firewall
 
 # RG-9: hardening baseline
 ProtectHome=yes
--- a/build/systemd/spinifex-firewall.service
+++ b/build/systemd/spinifex-firewall.service
@@ -1,0 +1,24 @@
+[Unit]
+Description=Spinifex host firewall (nftables input policy)
+Documentation=file:/etc/spinifex/firewall/spinifex.nft
+Wants=network-pre.target
+Before=network-pre.target shutdown.target
+DefaultDependencies=no
+ConditionPathExists=/etc/spinifex/firewall/peers.nft
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+StandardInput=null
+ProtectSystem=full
+ProtectHome=true
+ExecStart=/usr/local/lib/spinifex/spinifex-firewall-apply
+ExecReload=/usr/local/lib/spinifex/spinifex-firewall-apply
+TimeoutStopSec=10s
+
+# No ExecStop: flushing on stop would take vpcd's MASQUERADE and per-EIP FORWARD
+# rules with it. Remove the policy deliberately with
+# `nft delete table inet spinifex_filter`.
+
+[Install]
+WantedBy=sysinit.target
--- a/scripts/install-node.sh
+++ b/scripts/install-node.sh
@@ -37,6 +37,10 @@
 #                            Confirmed separately from --yes; unattended runs
 #                            must set SPX_WIPE_CONFIRM=wipe.
 #   --smoke                  Run smoke-test.sh --create-vpc on the first host
+#   --no-firewall            Do not enable the host firewall. The cluster ports
+#                            then stay reachable from anything that can route to
+#                            the node; only use it if the policy conflicts with
+#                            a site firewall that already scopes them.
 #   --yes                    Skip the confirmation prompt
 #   --dry-run                Print what would happen and touch nothing
 #
@@ -69,6 +73,7 @@ EMAIL=""
 PORT=4432
 TOKEN_TTL="30m"
 RUN_SMOKE=false
+INSTALL_FIREWALL=true
 WIPE=false
 ASSUME_YES=false
 DRY_RUN=false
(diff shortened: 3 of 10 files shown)


joshsiv-mulga:
## Multi-node verification on env1 (5 nodes)

Applied non-destructively to the running cluster, one node at a time with a health check between. env1 has no `br-lan`, so lan folds onto `br-vpc` — peers are the five WAN plus five 10.9.7.x addresses, encap the five 10.9.7.x.

**Encap mesh survives peer scoping.** All five nodes report 4 Geneve tunnels and 4 established IPsec SAs with the policy active. This is what env13 could not cover.

```
192.168.1.21 geneve=4 ipsec=4 fw=active
192.168.1.22 geneve=4 ipsec=4 fw=active
192.168.1.23 geneve=4 ipsec=4 fw=active
192.168.1.24 geneve=4 ipsec=4 fw=active
192.168.1.25 geneve=4 ipsec=4 fw=active
```

**Peer scoping is real, not vacuous.** From peer `10.9.7.4`, ports 4432 / 6642 / 4222 on `10.9.7.5` are OPEN. From a non-peer address, every node blocks 6641 / 6642 / 4222 / 4432 while 22 and 9999 still answer.

**Full-cluster function.** `smoke-test.sh --create-vpc --nodes 5` passes — five instances, one per physical node, each with an EIP and SSH-verified. That exercises cross-node scheduling over peer-scoped NATS, the `ime-*` rule on every node, and the EIP path.

## One operational note

Applying a stateful policy to an already-running node briefly drops flows that predate it. On the first node, peer OVN NB (6641) and SB RAFT (6644) data arrived as `ACK PSH` to established local ports and was dropped for a few minutes — those connections have no conntrack entry, so `ct state established` cannot match them. They recovered on their own without reconnecting (`nf_conntrack_tcp_loose` picks a mid-stream flow up once it sees traffic) and drops went to zero.

This is a property of enabling any stateful firewall on a live host, not of these rules, and the normal paths do not hit it: the unit orders before `network-pre.target`, so on boot the policy is in place before any connection exists, and at bootstrap the services have only just started. Worth knowing before applying by hand to a long-running node.

joshsiv-mulga:
## Update: the peer set is now reconciled by the daemon

The first cut wrote `peers.nft` from `install-node.sh` and from mulga's `cluster-bootstrap.yml`. Both are gone. `host.ReconcileFirewall` in `spinifex-daemon` owns it now.

**Why.** There are four formation paths — `install-node.sh`, the ansible `cluster-bootstrap` roles, `scripts/baremetal/bm-bootstrap.sh` and the ISO firstboot. The installer-side approach covered two of them, duplicated the same logic in each, and went stale the moment a node was added or replaced. Every path starts the daemon, so one implementation covers four and re-runs on every startup.

**The bug this surfaced.** A node's own `spinifex.toml` names its peers by their **wan address only** — remote `[nodes.X]` blocks carry `host = <wan ip>` and nothing else, and only the local node's block has lan addresses. A peer set built from config alone therefore contains exactly one lan address (the node's own) and drops every peer's NATS, OVSDB and formation traffic. On env1 the first reconcile produced `peers=6` and the drop log immediately caught a peer SYN from `10.9.7.3` to 6642 being denied.

The set is now the union of four sources, none optional:

| Plane | Source |
|---|---|
| wan | `nodes.X.host` / `advertise_ip` |
| lan | `routes` in the generated `/etc/spinifex/nats/nats.conf` — the only complete record of the lan plane a node holds locally |
| lan (db nodes) | the `ovn_nb_addr` / `ovn_sb_addr` remote lists |
| vpc | the OVN Southbound `Encap` table |

An empty encap answer is an error, not a narrower set.

**`ProtectSystem=full` reaches the sudo child.** The daemon's namespace makes `/etc` read-only for everything in the cgroup, so the root helper could not write `peers.nft` either (`install: cannot create regular file … Read-only file system`). Fixed with `-/etc/spinifex/firewall` in `ReadWritePaths`; the directory stays root-owned `0755`, so this grants the daemon no write access of its own, and the leading `-` keeps a node installed before the firewall stage from failing to start.

**Helper hardening.** `set-peers` and `disable` verbs. The address validator no longer swallows rejections in a pipeline subshell, checks octet *ranges* rather than just shape (nft rejects `10.0.0.999` and takes the whole transaction with it, which would leave a peer file that fails at every boot), and rolls the peer file back if the ruleset will not load.

### Verification

- **env13, three clean `make cluster-reset` runs.** The first looked like a pass and was not — the reconcile failed and the "active" policy was a table left loaded from an earlier run, because `nft` state outlives a teardown. The ansible report now reads the peer file *and* the table. After the fix: `peer sets reconciled peers=3 encap_peers=1`, all nine services `ok`, `describe-regions` answering.
- **env1, one node of five, non-destructively.** After adding the NATS-route and OVN-remote sources the same node produced `peers=10 encap_peers=5` — byte-for-byte the hand-written set the earlier 5-node run validated — and rejoined with all five nodes `Ready`, all services `ok`, and no peer-sourced drops.
- `make preflight` green.

Still blocked on the gold-image rebake for CI, as discussed — `provision.sh` carries `nftables` and `roles/install` installs it directly so existing envs are not stuck behind the image.

PR #771, opened 2026-08-12
feat: reconcile systemd unit files on upgrade

joshsiv-mulga:
## Summary

`spx admin upgrade` migrated config files only. Unit files were written once at install time by `install_systemd()` in `scripts/setup.sh` and never re-asserted, so a node upgraded by swapping the `spx` binary kept whatever units it was installed with. Any fix living in a unit rather than in code never reached it.

That includes the fixes merged in #766: `--unless-restarting` on `spinifex-shutdown.service` and `TimeoutStopSec=150` on `spinifex-viperblock.service` are both unit-file changes.

## Why this matters

Reproduced on a lab node. A `spinifex-viperblock.service` predating `a4274f95b` has no `KillMode=`, so it defaults to `control-group`:

| viperblock unit | qemu PID before → after | nbdkit before → after |
|---|---|---|
| `KillMode=process` | 2302543 → 2302543 | 2302417, 2302425 → both alive |
| `KillMode` absent | 2304606 → 2304606 | 2304496, 2304504 → **all gone** |

Viperblock deliberately spares nbdkit for the drain path, `TimeoutStopSec` then expires mid-flush, and systemd escalates to a cgroup-wide SIGKILL. Guest QEMU sits in the daemon's cgroup under its own `KillMode=process` and survives with no storage path.

## Approach

Units are verbatim and untemplated, so this is reconciliation, not the stepwise transformation the config registry performs.

- Each unit carries `# spinifex-unit-version: N` as its first line, mirroring the existing `# spinifex-config-version: N` marker. systemd ignores comments.
- `build/systemd/` stays the single source of truth. A generated `units_gen.go` embeds the contents, with an invariant test asserting it is byte-identical — `//go:embed` cannot reach outside its package, and putting an `embed.go` in `build/systemd/` would mean tightening globs in ten packaging consumers, where missing one is silent.
- `systemd.Reconcile(root, opts)` decides per unit:

| Installed | Action |
|---|---|
| Missing | Install |
| Marker version older, or absent (version 0) | Replace, after timestamped backup |
| Content identical | No-op |
| Marker version equal, content differs | Report as operator-modified, never overwrite |

Drop-ins under `<unit>.d/` are untouched, so `systemctl edit` overrides survive a replace.

- **Nothing is restarted.** `daemon-reload` alone makes `KillMode` and `TimeoutStopSec` apply to the next stop, so drift is corrected without disturbing a running guest.
- Adds `--dry-run`, `--units-only`, `--skip-units`.

## Verification

Live, on a node carrying the real drift:

| Check | Result |
|---|---|
| `--dry-run` reports 17 units `0 → 1 (stale)` | nothing written; `KillMode` still `control-group`, no backups |
| `--yes` replaces all 17 with backups | ✅ |
| `KillMode` | `control-group` → `process` |
| `TimeoutStopSec` | 30 → 150 |
| `spinifex-shutdown` ExecStop | `--only-if-host-stopping` → `--unless-restarting` |
| Service MainPIDs across upgrade | viperblock and daemon both unchanged — nothing restarted |

The node was then reverted to its original binary and units, confirmed by sha256 and unit content.

`make preflight` green.

**Not verified:** the guest-level confirmation — launch a guest, stop the target, observe nbdkit surviving — is outstanding. It was blocked by an unrelated AMI-resolution fault on the lab cluster (`mulga-leg2v`), under separate investigation. The KillMode causation is already established by the reproduction above, so this is confirmation of a demonstrated causal chain rather than the load-bearing evidence.

## Invariants

`spinifex/systemd/units_invariants_test.go` invariants are binding and stay green. Two additions: the generated copy must match `build/systemd`, and a unit body change requires a version bump.

One narrow change to `TestRG11_LeanUnits`: the version marker is excluded from the 12-line rationale-comment budget, since two units were already at exactly 12. `maxComments` is unchanged and no case was removed — the marker is structural metadata, not rationale prose.

Bead: `mulga-bzpgk`


joshsiv-mulga:
## Guest-level verification complete

The verification gap noted in the PR description is now closed. Same node, same stop command, drift recreated by removing `KillMode=process` from `spinifex-viperblock.service`.

**Phase 2 — buggy unit.** Guest on bottlebrush, qemu PID 2360848, nbdkit 2360731/2360741. `systemd-cgls` confirms the split that enables the bug: nbdkit in `spinifex-viperblock.service`'s cgroup, qemu in `spinifex-daemon.service`'s.

```
21:45:30 spinifex-viperblock.service: State 'final-sigterm' timed out. Killing.
21:45:30 spinifex-viperblock.service: Killing process 2360741 (nbdkit) with signal SIGKILL.
21:45:30 spinifex-viperblock.service: Failed with result 'timeout'.
```

qemu PID 2360848 unchanged — guest left running with no storage backend.

**Phase 3 — reconcile.** `--dry-run` listed all 17 units stale (0→1) and changed nothing. `--yes` replaced them with backups. `KillMode` → `process`, `TimeoutStopSec` → 150, ExecStop → `--unless-restarting`. MainPIDs before and after: viperblock 2361470/2361470, daemon 2361481/2361481 — nothing restarted.

**Phase 4 — fixed unit.** Guest qemu PID 2363675, nbdkit 2363569/2363577.

```
21:54:38 Draining local node "bottlebrush" (gate -> drain), timeout 2m0s/phase
21:54:38 QMP event POWERDOWN sent to guest
21:54:58 WARN Timeout waiting for PID file removal
21:54:58 [qemu-stderr] terminating on signal 15 from pid 2361481 (/usr/local/bin/spx)
21:54:58 INFO VM process exited cleanly
21:54:58 INFO Unmounted volume ... (os and efi)
21:54:58 [DRAIN] bottlebrush: OK
```

nbdkit: `Close, flushing block state to disk` → `VB Close` → chunk GC sweep → `NBDKit exited code=0` → `Deactivated successfully`. No SIGKILL and no timeout in either journal.

Drained gracefully, not killed. The drain completed inside its 120s budget, so systemd's `TimeoutStopSec` never fired.

| | Buggy unit | Reconciled unit |
|---|---|---|
| nbdkit | cgroup SIGKILL at TimeoutStopSec | clean exit, code 0 |
| qemu | left running, orphaned, no backend | SIGTERM after QMP powerdown, exited cleanly |
| viperblock.service | `Failed (Result: timeout)` | `Deactivated successfully` |

The node was reverted to its original binary and units afterwards — sha256 `524566915eefeb…` and `KillMode=process`/`TimeoutStopSec=30`/`--only-if-host-stopping` all match the pre-test baseline.

Two incidental observations, neither caused by this change: guest placement across three Ready nodes needed six launch cycles to land on a chosen node (consistent with round-robin scheduling, not a placement bug), and in Phase 2 the control plane still reported the instance `running` after its qemu had been killed out of band.

PR #779, opened 2026-08-12
fix(test): restore in-package tests that reach unexported symbols

joshsiv-mulga:
## Problem

`dev` does not compile:

```
vet: cmd/spinifex/cmd/admin_upgrade_test.go:47:20: undefined: validateUpgradeFlags
vet: spinifex/systemd/units_invariants_test.go:315:7: undefined: unitVersionRe
```

#776 moved both files to external test packages. In parallel, #771 and the admin-upgrade work added references to unexported symbols inside those same files. Both branches were cut before the other merged, so both were green in isolation and the merge result was not. Reverting those two files to in-package is the fix; the other six files #776 moved are unaffected.

## Knock-on: the systemd guards were silenced

Because `spinifex/systemd` did not compile, `TestGeneratedUnitsMatchSource` and `TestUnitBodyPinnedToVersion` stopped running, and drift landed unnoticed. `units_gen.go` was missing `spinifex-firewall.service` and carried a stale `spinifex-daemon.service`, both from #769. This PR regenerates it.

## Two things for follow-up

**1. `go generate ./...` cannot regenerate `units_gen.go`.** The directive lives in `spinifex/systemd/generate.go`, which is `//go:build ignore`, so `go generate` never scans the file. Every failure message tells you to run `go generate ./...`, which silently does nothing for this file. The generator has to be run directly from `spinifex/systemd`. Worth moving the directive to a non-ignored file in the package.

**2. `TestUnitBodyPinnedToVersion` still fails, and needs an owner decision.**

```
spinifex-daemon.service: body hash for version 1 changed — bump # spinifex-unit-version and add a new unitBodyHashes entry
spinifex-firewall.service: no entries in unitBodyHashes — add one for version 0
```

#769 changed `spinifex-daemon.service` but left it declaring version 1, so as it stands the change would not be rewritten onto already-deployed hosts. That is exactly what the guard exists to catch. Bumping to 2 is probably right, but it changes upgrade behaviour for a security feature, so I have left it to #769's author rather than guessing.

## Testing

`go vet ./...` clean. `spinifex/systemd` now compiles and `TestGeneratedUnitsMatchSource` passes.


joshsiv-mulga:
Fixed the version pin as well, so this PR now leaves `dev` fully green.

`Reconcile` replaces a unit only when the installed version is lower than the embedded one. Same version with different content is classified `ActionConflict` and deliberately left alone as operator-modified (`reconcile.go:135-141`). #769 changed `spinifex-daemon.service` — adding `-/etc/spinifex/firewall` to `ReadWritePaths`, which the daemon needs to write `peers.nft` — but left the marker at version 1. Every already-deployed host would have kept its old unit and the firewall would have silently not worked there.

- `spinifex-daemon.service` bumped to version 2, body hash pinned for 2 alongside the retained 1.
- `spinifex-firewall.service` had no version marker at all, so it read as version 0. Given a marker of 1 and its body hash pinned.

`make preflight` passes.

joshsiv-mulga:
Also fixed the generator wiring in this PR rather than leaving it as a note.

The `//go:generate` directive sat in `generate.go`, which is `//go:build ignore`. `go generate` does not scan files excluded by build constraints, so the directive never ran — while every drift failure message instructed the reader to run `go generate ./...`. That is how the `units_gen.go` drift, and the daemon unit version bug behind it, got in unnoticed.

Moved the directive to `reconcile.go`, which is part of the package and is scanned, and dropped the dead one from `generate.go`.

Verified by appending a marker to `units_gen.go`, running `go generate ./...`, and confirming the file came back byte-identical to before the corruption. That was a no-op prior to this change.

PR #872, opened 2026-08-25
feat(console): accept credential handoff from mulgadc.com signup

benduncan:
Replaces #868, which was closed. The quota work from that branch is already on `dev`, so what remains here is only the console credential handoff, re-applied cleanly on top of current `dev`.

## What this does

Lets the signup form on mulgadc.com drop a new customer straight into the console signed in, rather than asking them to copy a 40-character secret between tabs. The credentials arrive by `postMessage` and go directly into `exchangeForSession`, so the long-lived secret never touches a URL, the DOM or storage — only the short-lived STS session is persisted.

## Security model

Every inbound message is checked three ways: the origin is matched against a hardcoded allowlist, the source must be our `window.opener`, and the type must be the expected one. The source check is the one that matters most, since without it any frame on an allowed origin could ask for the credentials. The allowlist is never derived from the message, and the ready ping is posted to an explicit target origin rather than a wildcard.

The whole inbound payload is parsed with zod at the boundary, so nothing downstream branches on the shape of an untrusted value.

## Notes for review

- The ready ping re-announces on a short interval rather than firing once. The opener tab is backgrounded the moment it opens us, so its listener may not be attached when we mount; retrying until the credentials arrive is the difference between working in a test and working for a real customer.
- The login route's search schema accepts `handoff` as either `"1"` or `1`. The router coerces the param to a number, and matching only the string strips it from the URL, after which the handoff silently never runs.
- Degrades invisibly when unused: opened directly, or by an opener that does not implement this, no message arrives and the normal form stays usable.

## Testing

- `console-handoff.test.ts` covers the allowlist, the source check, the type check, a malformed payload, STS rejection, once-only behaviour and the no-opener case.
- Frontend: `pnpm fix` clean, `pnpm test` 1375 passing across 124 files, `pnpm build` regenerated the bundle.
- `make preflight` passes.


spinifex/services/spinifexui/frontend/src/routes/login.tsx
```diff
@@ -53,8 +66,26 @@ export const Route = createFileRoute("/login")({
...
-  const { reason } = Route.useSearch()
+  const { reason, handoff } = Route.useSearch()
   const [authError, setAuthError] = useState<string | null>(null)
+  // When mulgadc.com/signup opens us with ?handoff=1 it will postMessage the
+  // new account's credentials; receive them and sign in without the form. If
+  // nothing arrives (opened directly, or the exchange fails) we fall back to
+  // the form, so this is invisible when unused.
+  const [handingOff, setHandingOff] = useState(handoff === "1")
+  useEffect(() => {
+    if (handoff !== "1") {
```
juliansommer: can we not just make this if handoff === 1, return startconsolehandoff and otherwise don't do anything? then we can delete nohandoffcleanup + no cleanup in console-handoff.ts

PR #977, opened 2026-09-01
fix: pause a guest on a backend I/O error instead of handing it EIO

benduncan:
## Summary

A guest whose storage backend stopped answering was destroyed by it. Every
virtio-blk device was created with `werror=report,rerror=report`, so QEMU handed
the backend failure straight to the guest as `EIO`; ext4 aborts its journal on
that and remounts read-only, and the write that failed is simply gone. Under
`werror=stop` QEMU pauses the VM and **holds** the failed request, so a backend
that comes back resumes the guest with nothing lost.

This PR carries both halves: the e2e suite that reproduces the defect, and the
fix that closes it. Plan and evidence:
`docs/development/bugs/storage-fault-corrupts-guest-root-filesystems.md`
(Phases 2b, 2c and part of 3b). Bead `mulga-j0jdo`.

#966 (`fix/drain-seal-failure`) has landed and `dev` is merged in, so this diff
is now the storage-fault work alone.

One unrelated commit rides along because it blocks CI for anything branched off
`dev`: `TestDistributeInstancesSpread_BelowMinCountRollsBackAndReleases` and
`TestDistributeInstancesSpread_FinalizeFailureRollsBack` are both `t.Parallel()`
and both call `noopTerminateRetrySleep`, which swaps a package-level variable
production code reads. That is a real data race, not an ordering quirk, and the
race detector caught it here. Neither test is parallel now.

## What changed

- **`werror=stop,rerror=stop`** on all three call sites — the boot drive, the
  cold-boot data volume and the hot-plugged data volume.
- **`reconnect-delay=30`** on the NBD transport, so a storage process restarting
  on the same socket pauses requests rather than failing them. On the legacy
  `-drive` form it has to be `file.reconnect-delay`: the bare name reaches the
  raw format driver, which has no such option. Verified against a live
  `qemu-nbd -t` socket with controls in both directions.
- **Auto-resume in the QMP heartbeat.** Run state `io-error` is no longer
  mistaken for an unresponsive monitor; it issues `cont`, which re-submits the
  held request. After ten attempts the instance is reported impaired and the
  cadence drops to one attempt per ten heartbeats — but it never stops. QEMU
  holds the request for as long as the guest exists, so a backend that returns
  an hour later can still be resumed, and a five-minute timer that gave up would
  strand a guest that could have recovered.

## Reversal recorded, not quietly edited

`werror=report` was a deliberate choice, made so a full storage pool reports
`ENOSPC` to the guest instead of pausing the VM. That optimised for the guest
staying inspectable; the governing requirement is *never resume from an
incorrect state*, and the two select different answers. `stop` is the only
policy under which no write is lost or reordered. Proxmox defaults to `stop`
for the same stated reason. The `ENOSPC` path keeps its meaning at the
viperblock layer, which is the layer that can be selective about it.

## Results on dev-prod (3 nodes, RS(2,1))

Both reproductions were run red on the deployed cluster and green on the same
cluster with this change.

**`TestGuestSurvivesBackendOutage`** — predastore SIGSTOPped on 2 of 3 nodes for
7 minutes, load on the data volume *and* the root filesystem:

```
· no corruption signatures in the guest console or dmesg
· /mnt/fault is still mounted read-write
· sha256 verification passed: every acknowledged write survived
```

**`TestGuestSurvivesNbdkitLoss/Killed`** — the volume's own nbdkit killed, held
60s, then put back on the same socket. Red beforehand with the signature the
suite exists to catch:

```
the guest was shown I/O errors and reacted to them:
  EXT4-fs error, Aborting journal, Remounting filesystem read-only, I/O error, dev vdb
```

Green after, with the same three assertions as above.

The observation loop is where the change is legible: under `report` the guest
stayed `guest_ssh=reachable` throughout and used that reachability to abort its
journal; under `stop` it reports `guest_ssh=unreachable instance_state=running`
for the whole outage — paused, not damaged — and answers again once `cont`
replays the held request.

## Test-side changes

- **The killed-nbdkit fault**, which is far narrower than freezing a
  cluster-wide service: one volume on one host, relaunched afterwards, because a
  volume left dead tests the fault and not the recovery.
- **`dd` and `sha256sum` replace `fio`.** The suite needs real in-flight I/O and
  a byte-exact integrity check, not a benchmark, and installing a package over
  the internet inside the guest was by some distance the least reliable step in
  the run.
- **One guest per test.** Every fault here is meant to hurt, and a guest that
  survived one is not a clean start for the next — after the frozen variant the
  guest was left with `jbd2/vdb-8 blocked for more than 245 seconds` and the
  killed variant then failed on damage it had not caused. `InstanceSpec.Scope`
  reaches the memo key so each test gets its own.
- **Two relaunch defects that made a working fix look broken.** The socket
  liveness poll ran unprivileged against a `0770` directory and could never
  succeed; and the relaunched socket was left `0755`, where QEMU — a different
  user in the same group — needs group write, so the volume came back serving a
  socket nothing could reach. The readiness check is now
  `sudo -u spinifex-daemon test -w`, the property that matters rather than a
  proxy for it.

## Not covered here

Running both `TestGuestSurvivesNbdkitLoss` variants in one pass is blocked on
`mulga-gblvl`: a second concurrent guest's EIP ingress does not converge inside
the 5-minute SSH wait, because the reconciler sits in an 8-minute backoff held
there by unrelated stopped instances' ports. Unrelated to storage; the variants
were run one at a time.

---

## Update: the suite is green, and what it took

Six further commits. The first three are the defects the suite found once it was
reliable enough to be believed; the last three are the suite itself.

### The guest was not being resumed for long enough (`vm/lifecycle.go`)

`resumeAfterIOError` gave up issuing `cont` after `qmpMaxIOErrorResumes` (10) and
from then on only logged. The comment above it said the opposite — that it never
stops — but the early `return` skipped the resume. So a guest whose backend
returned after ten heartbeats stayed paused forever. This is the change that
actually made `TestGuestSurvivesBackendOutage` pass: in one run predastore was
proved serving through all three gates while `i-c1ea2deae37310a00` sat at
`attempt 50`, then `attempt 60`, exactly five minutes apart.

The retry now runs at heartbeat rate for as long as the guest exists. Only the
**logging** is throttled, to one line in ten once impaired — a retry costs one
replayed I/O, and a guest that waited minutes to learn its backend had returned
was paying far more for it than the retry ever saved.

### An unadoptable nbdkit export was leaked (`services/viperblockd/`, `mulga-wx0rb`)

Restart recovery adopts the boot volume, then fails the `-efi` one with
`claim lease: context deadline exceeded` — JetStream is still catching up at
daemon start and the KV claim uses the client's 5s default. Recovery gave up on
the first error and never retried, so the EFI volume had no registry entry; at
terminate the unmount returned `NotFound` and killed nothing, and the export
outlived the instance.

Two changes, both narrow:

- `rebuildMountedVolume` retries `buildVB` 4 times with a 3s backoff, so a claim
  that fails only because the KV is cold succeeds on the next attempt.
- `unmountVolume`'s `not found` branch calls `reapUnregisteredNbdkit`, which
  reaps a **corroborated** export for that volume — plugin suffix, base dir and
  live endpoint all matching — by delegating to `reapOrphanedNbdkit`, which
  already declines anything a live process still references. It reports
  `Reaped` in the response rather than reaping silently.

Verified: a run that previously ended with two stray `-efi` nbdkits across the
three hosts now ends with zero.

### predastore's admin port is scoped to the cluster plane (`scripts/setup.sh`)

`/healthz` and `/readyz` are unauthenticated and name which peers a gate cannot
reach, so 8660 goes in the existing peer-scoped nft rule beside NATS and the OVN
databases — never the WAN. One port added to one existing rule, plus the port
table in `docs/security/network-connections/README.md`. **Not applied to any
running host**; it takes effect the next time `setup.sh` runs.

### The harness now asserts the write path, not just a read

Under RS(2,1) a read reconstructs from one surviving shard while a write needs
two durable, so a single successful round trip through a gate says nothing about
whether a guest can be served. `AssertPredastoreServing` now holds for three
rounds ten seconds apart, restarting the count on any failure, and then calls
`AssertPredastoreReady`, which decodes `/readyz` and reports every check that is
not `ok` by name. The matching predastore-side change — gating `blob_nodes` on
the write floor and naming the unreachable node — is mulgadc/predastore#141.

Caveat, stated rather than papered over: dev-prod's deployed `predastore.toml`
predates `admin_port` in the template, so there is no admin listener and the
readiness assertion logs and skips on every host. It is unit-tested but not yet
proved against a live cluster.

### Two harness defects that cost whole runs

- **A duplicate key pair name failed every guest in a run.** One `CreateKeyPair`
  produced two pairs with the same name, different fingerprints and different
  `KeyPairId`s, 1 ms apart. `RunInstances` resolved the name to one of them while
  the harness held the other's private key, so all four guests were rejected at
  the SSH gate after five minutes each and the run looked like a storage failure.
  `EnsureKeyPair` now asserts the name resolves to exactly one pair and fails
  loudly if not. The underlying P1 is filed as `mulga-4f6r4` with the evidence;
  it is not fixed here.
- **Teardown failed on a volume it had already released.** Terminate releases the
  volume, so the follow-up `DetachVolume` returns `IncorrectState` and failed a
  test whose assertions had all passed. That one code is now tolerated and the
  wait for `available` proceeds; `VolumeInUse` and non-AWS errors are not.

## Result

Full suite on dev-prod, all four tests in one pass:

```
--- PASS: TestGuestSurvivesBackendOutage (1073.87s)
--- PASS: TestGuestSurvivesNbdkitLoss (1386.61s)
    --- PASS: TestGuestSurvivesNbdkitLoss/Frozen (560.20s)
    --- PASS: TestGuestSurvivesNbdkitLoss/Killed (826.42s)
--- PASS: TestStopStartPreservesWritesOnSameNode (605.87s)
--- PASS: TestCrossNodeStartAfterFailedSealTakesOverLoudly (617.77s)
ok  github.com/mulgadc/spinifex/tests/e2e/storagefault 3767.792s
```

Both `TestGuestSurvivesNbdkitLoss` variants now run in the same pass, which the
"Not covered here" section above said they could not — `mulga-gblvl` did not
recur once each test took its own guest. Teardown left no stray nbdkit belonging
to any instance the run launched.

## Not covered here

- **Network-partition fault injection**, deliberately deferred.
- **`mulga-h2z6m`** (the predastore gate re-dial wedge) **did not reproduce** in
  any of the three full runs. Rather than close it on absence, a real recovery
  hole with the same shape was fixed in the connection pool — an evicted, empty
  slot refusing the connection its peer was offering — with a negative control
  proving the tests catch it. See mulgadc/predastore#141. The bead stays open.

## Review round (2026-09-04)

Josh's four comments, each answered inline on the thread it was raised in.

- **`otelsetup/metrics.go` — type and const block** (25fcf5d5). `FenceOutcome` is a string type, the four values are one `const` block, and `RecordVolumeFence` and `fenceVolume` both take that type instead of a bare `string` — so the closed domain the comments were asserting is now enforced by the compiler.
- **`daemon` importing `viperblockd`** (b4f7e1fe). New leaf package `spinifex/services/viperblockd/vbwire` holds the wire contract: `VolumeFencedSubject`, `VolumeFencedEvent`, the dirty bucket name and record shape, `BindDirtyBucket`, `UnsealedVolume` and `ListUnsealedVolumes`. It imports nothing from spinifex. Those two symbols were the only reason `daemon` imported `viperblockd`, so the edge is gone rather than narrowed; `cmd/spinifex/cmd/admin.go` drops it too, and `cmd/service.go` keeps its import legitimately to construct the service.
- **Infra reference in `StartInstances.go`** (c5d2669a). The comment named a dev-prod host; it now says "the name of the node holding the volume". Two test fixtures this branch added used the same hostname and now use `node-b`.

## storagefault now has a nightly cell, on three nodes

I claimed earlier in this PR that the gate already ran this suite on both topologies. That was wrong. `storagefault` is *eligible* for both — it is in `E2E_SUITES_SINGLE` and `E2E_SUITES_MULTI` — but eligibility is not scheduling: `e2e.yml`'s default `SUITES` list never names it, so **it has never run in CI at all**, on either topology. The only passes it has are the manual dev-prod runs quoted above.

Nor should it run on the gate. A full pass is ~63 minutes on dev-prod bare metal and almost all of that is deliberate waiting — one test holds predastore SIGSTOPped for seven minutes to prove a guest survives it. The gate currently finishes well inside its budget and adding an hour to it is the wrong trade.

So it gets a nightly cell, following eks/ecs/rds. Cell 29 is **real-multi on `ci-multi`**, not single-node: single is RS(1,0), where the shard floor cannot be broken and a cross-node takeover after a failed seal does not exist — the two assertions worth a dedicated cell are the two a single node cannot make. It cannot join a permutation cell either, because it SIGSTOPs a cluster-wide service and every suite sharing that environment would report the outage as its own failure.

Three consequences in `e2e-nightly.yml`: a per-cell `timeout_minutes` defaulting to the previous 50, since a 90-minute suite cannot run in a 50-minute job; the permutation steps gated on `matrix.suite == ''` so dedicated cells do not also run them; and a `Run Multi Suite E2E` step that builds `NODES_CSV` from the tofu inventory and asserts the suite against `E2E_SUITES_MULTI` first. The 90m budget is declared once, in `e2e_suite_timeout`.

## Closing run on dev-prod (2026-09-04) — and the bug it found first

The confirming run **failed**, and the cause had nothing to do with storage. Guests refused SSH with `Permission denied (publickey)`, and the failure correlated exactly with the node: 4 of 4 guests on casuarina failed, 3 of 3 on the other two passed.

vpcd serves IMDS, and `NewIMDSServiceImpl` opens five JetStream KV buckets at construction under three different failure policies — `spinifex-vpc-enis` fatal, and `spinifex-instance-state`, `spinifex-vpc-security-groups`, `spinifex-vpc-subnets` and `spinifex-vpc-vpcs` all silently and *permanently* degrading on a ~5s timeout. That inconsistency is the defect. On a whole-cluster restart every service starts at once and races JetStream electing a leader, so all three nodes logged the instance-state bucket as unavailable. Bottlebrush and ironbark were then saved by accident: seconds later they also failed the ENI open, which *is* fatal, so systemd restarted them into a healthy JetStream. Casuarina's ENI open happened to succeed, so it never crashed, reported `NRestarts: 0`, and ran 2.5 hours with its instance-record loader nil.

IMDS looks up an instance locally first, so that alone was survivable — except the daemon wrote `instance-state.json` `0600 spinifex-daemon:spinifex` while vpcd runs as `spinifex-vpcd`. With both sources gone every instance-backed leaf 404'd. cloud-init crawls directories before leaves, so the ENI-backed ones answered and the crawl died on the first leaf, fell back to `DataSourceNone`, and wrote no `authorized_keys`. The guest booted keyless with nothing in any log calling it an error.

Two fixes, in `9b335740`:

- `kvstore.OpenWithRetry` retries a bucket open for 90s before the caller's policy applies, and all five opens go through it. Failing the instance-record open is now fatal by design rather than a warning.
- `daemon.LocalStateFileMode` pins the state file to `0640` and chmods explicitly, because `OpenFile`'s mode is masked by whatever umask the service inherits.

Proven on redeploy — all three nodes hit the same race, retried, and opened on attempt 2 after ~7s:

```
WARN  kvstore: not ready, retrying   what=spinifex-instance-state attempt=1 window_ms=90000
INFO  kvstore: opened after retrying what=spinifex-instance-state attempts=2 waited_ms=7010
INFO  IMDS service initialized       eni_bucket=spinifex-vpc-enis
```

Ruled out with evidence first, each of which looked plausible: a stale EIP/MAC binding (OVN answers a NAT external IP with the guest's own MAC, so an ARP entry proves nothing either way), duplicate NB NAT rows, security groups, a missing AMI, the network path, IMDS API-version handling, and incomplete KV records.

Beads `mulga-gtpzz` (P1) and `mulga-dhqxr` (P2) closed. **`mulga-mm2oc` (P1) stays open and is not fixed here:** IMDS serves an instance it cannot resolve as "this instance has no key pair". Removing this trigger does not remove that surface.

## Re-run: four of five green, and the fifth is a harness gap

```
--- PASS: TestGuestSurvivesBackendOutage (1164.20s)
--- PASS: TestGuestSurvivesNbdkitLoss (1399.56s)
    --- PASS: TestGuestSurvivesNbdkitLoss/Frozen (563.17s)
    --- PASS: TestGuestSurvivesNbdkitLoss/Killed (836.39s)
--- PASS: TestStopStartPreservesWritesOnSameNode (581.65s)
--- FAIL: TestCrossNodeStartAfterFailedSealTakesOverLoudly (634.39s)
```

Both substantive assertions of the failing case passed — the instance started on another node while its own host was down, and that node logged the takeover naming the previous holder. It then died on the read-back with `sudo: A terminal is required to authenticate`. One line below the call site the test tolerates the pattern having changed, calling it "the documented cost of starting from the backend checkpoint", but `readPattern` `t.Fatalf`s on any exec error, so it cannot tell "read a different value" from "could not read the device". The guest booted and took SSH because `authorized_keys` is on the root filesystem from first boot; `/etc/sudoers.d/90-cloud-init-users` was not in the backend checkpoint, so passwordless sudo rolled back with the rest of the root filesystem. The suite accepts the volume rolling back but not its own access path rolling back with it. Bead `mulga-ho0gz`.

Stated plainly: this test asserts nothing about *how much* is lost in either direction, so the run is not evidence that the amount rolled back was correct.

## Benchmark

Post-deploy run on `v1.18.0-190-g9b335740`, same harness and instance type as the two runs before it. Guest disk cluster means, read IOPS: 4k 6,664 (+0.9% on the 08-31 baseline), 16k 5,112 (+3.1%), 128k 1,425 (+7.6%), 1M 199 (−4.3%). Host control is unchanged, so the guest figures are readable.

The one negative cell is a single node: 1M reads 230 / 226 / 140 per node, and casuarina's 140 carries a p99 of 2,466 ms against 750 and 759. Excluding it, 1M is 228 — `+9.6%`. Casuarina is the node the IMDS race had left degraded earlier the same day and is also the NATS leader, so this reads as a single-node artifact rather than a 1M regression, and is worth re-measuring on a quiet cluster. Network is 0.96 Gbit/s against 0.97, unchanged.

## Also folded in

- **`5ada1a02` — scope the `ime-*` firewall accept to IMDS and VPC DNS only.** The per-tap OVS demux delivers any guest packet addressed to `169.254.169.254`/`.253` onto an `ime-*` host port regardless of L4 port, and `iifname "ime-*" accept` admitted all of it, so every wildcard-bound host service answered guest packets on those addresses. **Not deployed to dev-prod and not exercised by the run above**, which ran on `9b335740`.
- **`abaf568b`** — docs for the instance-to-host plane.

---

## Green, five of five (`cd6b6eee` on dev-prod)

```
--- PASS: TestGuestSurvivesBackendOutage (702.43s)
--- PASS: TestGuestSurvivesNbdkitLoss (1387.81s)
    --- PASS: TestGuestSurvivesNbdkitLoss/Frozen (557.54s)
    --- PASS: TestGuestSurvivesNbdkitLoss/Killed (830.27s)
--- PASS: TestStopStartPreservesWritesOnSameNode (597.29s)
--- PASS: TestCrossNodeStartAfterFailedSealTakesOverLoudly (600.91s)
ok  github.com/mulgadc/spinifex/tests/e2e/storagefault 3329.046s
```

### `cd6b6eee` — the meta-data listing is no longer a guess

Every omission from the meta-data root is a positive claim about the instance. `public-keys/` absent means "launched without a key pair", and cloud-init believes it, writes no `authorized_keys` and boots the guest keyless. Resolving the instance was nevertheless best-effort: a backend error or a miss shortened the listing rather than failing it, and nothing recorded that the answer was a guess. That is `mulga-mm2oc`, and it is the surface the earlier vpcd race fell through — the race supplied the failed resolve, this supplied the silence.

It now withholds the listing with a **500**. cloud-init retries a 5xx and cannot retry a 200 that is merely wrong, so a permanent silent failure becomes a transient loud one. `servePublicKeys` already had exactly this policy for the leaf; the root contradicted it.

A profile lookup that fails still serves the listing minus `iam/`, and now logs it. The key pair is the field a guest cannot recover from losing, whereas a missing role surfaces through the credentials leaf, which reports the miss itself.

### The harness no longer fails on its own access path

`readPattern` needs sudo, and a volume opened from the backend checkpoint may predate cloud-init writing `/etc/sudoers.d/90-cloud-init-users` — indistinguishable from an unreadable device at the `dd` call site, which is what failed the previous run. The test now probes `sudo -n true` first, and when sudo is gone it reads cloud-init's recorded datasource to tell a rolled-back root filesystem from IMDS never having answered, **failing loudly on the latter** rather than banking it as storage loss.

**What this run does not prove.** The cross-node case passed on the ordinary path — `the pattern also survived the takeover: node1 had sealed enough of it` — so sudo worked and the new branch never executed. It is still unproven against a live rollback, and what differs between runs is how much the node sealed before the freeze, which the test does not control. `mulga-ho0gz` stays open for that reason.

### Verification before the run

- Deployed `cd6b6eee` to all three dev-prod nodes; all `Ready`, `IMDS service initialized` on each, and each hit the JetStream race and recovered via `OpenWithRetry` on attempt 2.
- Smoke-tested a 3-guest workload first: all three took SSH keys through IMDS under the new 500 behaviour, so the change does not break ordinary boot.
- **`5ada1a02` was not exercised by this run**, and did not need to be. The dev-prod host firewall is off by design and the nft ruleset is empty, so the `ime-*` scoping shipped in this deploy without any rule from it loading. It has since been verified on prod, where the firewall is armed.

`mulga-mm2oc` closed. `mulga-ho0gz` open.


spinifex/otelsetup/metrics.go
```diff
@@ -23,13 +23,42 @@ const actionAttrKey = "rpc.method"
 // IDs — those belong in the accompanying log line.
 const leakKindAttrKey = "resource.kind"
 
+// fenceOutcomeAttrKey names what a node did after losing a volume lease.
```
joshsiv-mulga: So many comments. And could have grouped these in a block declaration... and added a type:

```go
type FenceOutcome string

const fenceOutcomeAttrKey = "outcome"

const (
           FenceOutcomeTaken FenceOutcome = "taken"
           FenceOutcomeExpired FenceOutcome = "expired"
           FenceOutcomeStalled FenceOutcome = "stalled"
           FenceOutcomeKillFailed FenceOutcome = "kill_failed"
)
```

spinifex/daemon/daemon.go
```diff
@@ -69,6 +69,7 @@ import (
 	"github.com/mulgadc/spinifex/spinifex/objectstore"
 	"github.com/mulgadc/spinifex/spinifex/otelsetup"
 	"github.com/mulgadc/spinifex/spinifex/preflight"
+	"github.com/mulgadc/spinifex/spinifex/services/viperblockd"
```
joshsiv-mulga: Reintroducing viperblockd into daemon is a regression on the decoupling I have just done. Would be better to put `VolumeFencedSubject`, `VolumeFencedEvent` (and `ListUnsealedVolumes` for `admin.go`) in a leaf package like `spinifex/services/viperblockd/vbwire`.

spinifex/daemon/daemon_handlers_fence.go
```diff
@@ -0,0 +1,45 @@
+package daemon
+
+import (
+	"encoding/json"
+	"log/slog"
+
+	"github.com/mulgadc/spinifex/spinifex/services/viperblockd"
```
joshsiv-mulga: Same as above.

spinifex/gateway/ec2/instance/StartInstances.go
```diff
@@ -132,9 +132,27 @@ func startStoppedInstance(ctx context.Context, natsConn *nats.Conn, instanceID,
...
+		return nil, startRefusalError(responseError)
 	}
 
 	slog.InfoContext(ctx, "StartInstances: Command sent successfully", "instance_id", instanceID, "response", string(msg.Data))
 	return newStateChange(instanceID, 0, "pending", 80, "stopped"), nil
 }
+
+// startRefusalError rebuilds the daemon's error, keeping the message it sent.
+// Reconstructing from the code alone renders the generic ErrorLookup text
+// instead, which turns "bottlebrush holds this volume" into boilerplate about
```
joshsiv-mulga: Remove specific infra references

spinifex/otelsetup/metrics.go
benduncan: Fair on both counts — done in 25fcf5d5.

`FenceOutcome` is a string type now, the four values are one `const` block with a line or two each instead of a paragraph, and `RecordVolumeFence` takes the type rather than a bare `string`. `fenceVolume`'s parameter is typed the same way, so an arbitrary string can no longer reach the metric attribute — the compiler enforces the closed domain the comment was asserting.

spinifex/daemon/daemon.go
benduncan: Agreed, and your suggestion is what I did — b4f7e1fe.

New leaf package `spinifex/services/viperblockd/vbwire` holds the wire contract: `VolumeFencedSubject`, `VolumeFencedEvent`, the dirty bucket name and record shape, `BindDirtyBucket`, `UnsealedVolume` and `ListUnsealedVolumes`. It imports nothing from spinifex.

Those two symbols were the **only** reason `daemon` imported `viperblockd`, so the edge is removed rather than narrowed. `cmd/spinifex/cmd/admin.go` drops it too. `cmd/service.go` keeps its import, which is genuine — it constructs the service from `viperblockd.Config`.

spinifex/daemon/daemon_handlers_fence.go
benduncan: Same change — b4f7e1fe. This file now imports `vbwire` for `VolumeFencedEvent` and nothing else.

spinifex/gateway/ec2/instance/StartInstances.go
benduncan: Done in c5d2669a. Reworded to "the name of the node holding the volume", which is the fact the sentence was actually about.

The same hostname was in two test fixtures this branch added — `volume_held_elsewhere_test.go` and the fenced-subject test — both now use `node-b`.

PR #1034, opened 2026-09-15
fix: describe an EC2 instance well enough to plan against

joshsiv-mulga:
Six defects in the EC2 describe path, found by running unmodified Terraform Registry modules against env19 and planning immediately after a clean apply. They share one shape: the write succeeds, the read-back is incomplete, and the caller cannot tell.

The symptom that started it, on the live demo 01 stack:

```
# module.bastion.aws_instance.this[0] must be replaced
    ~ associate_public_ip_address = false -> true # forces replacement
Plan: 2 to add, 2 to change, 1 to destroy
```

The instance has a working public address. The provider reads `false` because the NIC carries no `Association` block, and that attribute is force-new.

## Beads closed by this PR

None yet. See the verification section below: all six are implemented and unit-tested but only one is verified live, so they stay open pending the rest.

| Bead | Defect |
|---|---|
| `mulga-ifscc` (P0) | NIC has no `Association`, so `associate_public_ip_address` reads false and forces replacement |
| `mulga-3s0s6` (P2) | `EnclaveOptions` never populated, so every plan carries a phantom `enclave_options` block |
| `mulga-bpvko` (P1) | Empty Reservations emitted for fully filtered-out groups, and one Reservation minted per node instead of per launch |
| `mulga-92uy6` (P1) | An incomplete fan-out returns an empty 200 that reads as authoritative |
| `mulga-fr4jp` (P2) | `DescribeVpcs` rejects the AWS filter spellings `cidr` and `isDefault` |
| `mulga-gppj5` (P2) | `describe-images --owners self` resolves the owner alias after filtering on it, so it matches nothing |

## Notes for review

**The reservation ID now rides in a NATS header.** `PrepareRunInstances` minted one per node, so a launch spread across three nodes came back as three reservations and no amount of merging in the gateway could undo it. The gateway mints the ID once and passes it as `X-Reservation-Id`; a node that does not read the header mints its own exactly as before, so an old node behind a new gateway and a new node behind an old gateway both behave as they do today. The daemon applies the override before the response is marshalled and before the VM state is written, and `PrepareRunInstances` assigns the same `*ec2.Reservation` to every instance, so the response and the persisted state cannot disagree.

**`distributeInstancesSpread` had the same defect** and is fixed the same way. It was outside the original diagnosis, but leaving it would have meant a spread placement group still fragmented into one reservation per node.

**Five existing tests encoded the `mulga-92uy6` defect** and now assert the 503 instead of a silent success. The safety property they were written to protect is "never a false `InvalidInstanceID.NotFound`", not "never an error", and that property is preserved. Two tests were deliberately left untouched: `TestDescribeInstancesChecked_FilterOnlyNoMatch_ReturnsEmptyNotNotFound`, because a filters-only query with no matches is a legitimate empty result, and `TestDescribeInstances_ExplicitIDNotFound_StaysSilent`, because the lenient variant is what the EKS and RDS reconcilers call.

**The `DescribeVpcs` filter names are added, not renamed.** `handlers/rds/network.go` hardcodes the old `is-default` spelling, and `handlers/rds/network_test.go` carries a comment recording that the camelCase spelling was tried and rejected. Both spellings work now.

**`cidr-block-association.cidr-block` can only ever match the primary CIDR.** `VPCRecord` holds a single `CidrBlock` and the association set is synthesised per request. Conformant-shaped, not conformant-complete; multi-CIDR is separate work.

**`IpOwnerId` is set to the instance's account ID.** Real AWS reports `amazon` for an auto-assigned public IP and the account ID only for an Elastic IP. Nothing in the codebase models that distinction today, so it was not invented here.

**One unrelated change:** `testutil/segscanoracle` now builds with `-buildvcs=false`. It fails on current toolchains regardless of this PR (VCS stamping errors when building a synthetic module under a temp dir) and it blocks `make preflight` for everyone. Reproduced identically on an untouched checkout of `dev`.

## Verification

`make preflight` passes. New regression tests cover every bead.

Live verification is **partial**, and the reason is not this branch. `make cluster-deploy ENV=env19` halted on the first node because predastore crash-loops on startup with a nil pointer dereference in `meta.(*Server).LeaderKnown`, filed as `mulga-1n8kt` (P0). That panic comes from the predastore pin in 442444a27 and reproduces without any of this work. The remaining two nodes therefore still run a 2026-09-10 build.

Against the one node running this branch, `TFC-0003` flips from `present` to `absent` — the empty-Reservation leak is confirmed fixed on live. The other four probes still read `present` because the instance, VPC and image paths are served by the nodes that could not be upgraded. They need a re-run once `mulga-1n8kt` is fixed and the cluster can take a full rolling deploy.


joshsiv-mulga:
Live verification is now complete on env19, and one claim in the description needs correcting.

## All six beads verified live and closed

`make cluster-reset ENV=env19` rebuilt all three nodes on this branch. That also cleared the predastore crash: a clean bootstrap on the same pin came up first time with `NRestarts=0`, so `mulga-1n8kt` only triggers when restarting into already-degraded raft state, not on every startup. The cluster had been logging "0 of 3 blob nodes answered" since 2026-09-10, well before any deploy touched it.

Five conformance probes moved `present` to `absent`:

```
absent   P0  TFC-0003  DescribeInstances returns an empty reservation ...   <- FIXED since baseline
absent   P0  TFC-0004  A NIC with a public address returns no Association    <- FIXED since baseline
absent   P2  TFC-0005  DescribeInstances omits EnclaveOptions                <- FIXED since baseline
absent   P2  TFC-0007  DescribeVpcs rejects a cidr filter                    <- FIXED since baseline
absent   P2  TFC-0008  DescribeImages ignores the self owner alias           <- FIXED since baseline
```

`mulga-92uy6` has no probe, so it was staged directly by stopping `spinifex-daemon` on one node:

| Scenario | Result |
|---|---|
| All nodes up, nonexistent ID | `InvalidInstanceID.NotFound` — unchanged |
| One node down, nonexistent ID | `ServiceUnavailable: cannot confirm status of instance(s) i-0000000000000dead: describe fan-out did not complete`, and the SDK retried it |
| One node down, real ID | Still resolves |

A two-instance `run-instances` returned a single reservation, `r-c0c37cb7bebf1f2c4`.

The full cheap probe sweep exits 0, so nothing that previously worked regressed.

## Correction

The description says the `DescribeVpcs` parse error "also names the offending filter to the client instead of only the log". **The handler-side half is real, but it does not reach the client yet**, so treat that sentence as describing intent rather than observed behaviour.

The message travels correctly over NATS and the gateway logs it — `InvalidParameterValue: The filter 'zzz-probe-marker' is invalid` — but the client still receives the canned text. `resolveErrorDetail` in `awserrors` only returns a message when the error is a `*codedError`. An error arriving over NATS is a `*decodedNATSError`, whose `Error()` is not an `ErrorLookup` key, so it unwraps to a plain `errors.New(code)`, the `AsType` check fails, and the message comes back empty. `decodedNATSError.message` is never read.

That is gateway-wide, affects every daemon-served action, and is the root cause of `mulga-p46w3` / TFC-0002 rather than anything in this cluster. Full analysis is on that bead. Nothing here needs changing for it — the `awserrors.Errorf` call is a prerequisite that becomes visible once that bead lands.

It also produced a false positive worth flagging: TFC-0002's probe provoked its error with `--filters Name=cidr`, which this PR makes valid, so the probe briefly reported that gap as fixed without testing anything. The probe now uses a filter name nothing will implement and correctly reads `present` again.

PR #1036, opened 2026-09-15
fix: carry a handler's refusal message to the client without leaking wrapper context

joshsiv-mulga:
A daemon that takes the trouble to explain a refusal had that explanation discarded one step short of the client, so every NATS-backed error surfaced as the generic `ErrorLookup` blurb for its code. The message crossed NATS intact all along — the wire format has always carried `Message` alongside `Code`. `resolveErrorDetail` then dropped it, because `decodedNATSError` was not the shape it recognises: its `Error()` returned the message and its `Unwrap()` returned a bare `errors.New(code)`, so the walk stepped past the message and returned an empty one.

## The part that needs review

Restoring the message is not safe on its own. Both encode sites sent `err.Error()`, which renders every `%w` layer including context a call site added purely for its own logs — a conformance probe observed `dial predastore 10.0.1.5:8443: connection refused` going onto the wire. Today that text is decoded and then thrown away, so it never reaches a client. The moment the decode side stops throwing it away, it would.

So this sends only `ResolveErrorDetail`'s curated message and adds **no** `err.Error()` fallback. `awserrors.Errorf` records its message at construction from its own format string, so an outer `fmt.Errorf("context: %w", inner)` wrapper is not part of it. What reaches the client is exactly what a call site chose to say.

Two existing tests pinned the opposite contract — `TestServeNATSRequest_WrappedErrorPreservesCodeAndMessage` asserted `err.Error() == "launch on node-1: InsufficientAddressCapacity"`, putting wrapper context on the wire, while `gateway_test.go:286` asserted that same `"launch on node-1"` must never reach the client. Both passed only because the gateway discarded the message. They are rewritten to pin the sanitized contract. `TestErrorHandler_WrappedErrorCode` itself is **unmodified** and still passes; it is the regression test for this risk.

## Consequence worth knowing

Uncoded handler errors (a bare `fmt.Errorf`, no registered code) no longer put their message on the wire. There are ~1292 such sites. This is **not** a client-visible change — the gateway already discarded those messages — but the gateway's `slog.Debug` line for them now shows the code alone. The daemon still logs the full reason at origin. `handlers/acm/renewal.go` is converted as the template for migrating one when its reason is worth surfacing.

## Also in here

`InvalidParameterValue` now names the rejected parameter. `ParseFilters` spelled the code into its own message text, so once the code travelled separately the client saw it twice; it now carries the code via `awserrors.Errorf` and its 25 callers return it unwrapped. All 25 already mapped it to `InvalidParameterValue`.

## Beads

- `mulga-p46w3` (P1) — `InvalidParameterValue` names no parameter. Fixed.
- `mulga-ky2zy` (P2, Julian) — handler diagnostics dropped at the gateway. Root cause fixed here; left open for its owner.
- `mulga-e9qqx` (P1, Julian) — daemon messages dropped on the NATS hop. Root cause fixed here, but its proposed `err.Error()` fallback is deliberately **not** implemented, being the disclosure path above. Left open for its owner.

## Verification

`make preflight` passes. Live on env19, before and after:

```
# before
DescribeVpcs: A value specified in a parameter is not valid, is unsupported, or cannot be used...
CreateDBInstance: Indicates an incorrect combination of parameters, or a missing parameter...

# after
DescribeVpcs: The filter 'zzz-not-a-real-filter' is invalid
DescribeNetworkInterfaces: The filter 'bogus-filter' is invalid
CreateDBInstance: DB parameter group default.postgres18 is of family postgres18, which cannot be used by a mariadb DB instance; it requires a group of family mariadb11.8
```

The RDS case is `ky2zy`'s exact reproduction, and its message is written daemon-side at `handlers/rds/parametergroup.go:522`, so it genuinely crosses the hop. Conformance probe TFC-0002 moved `present` to `absent`; a `--cheap-only` sweep exits 0 with nothing regressed.


joshsiv-mulga:
Added `3d683fb0d` after re-triaging older beads against this branch.

`ipPermissionsToSGRules` composes a real reason — `"IPv6 rules are not supported"` — and both `AuthorizeSecurityGroup{Ingress,Egress}` call sites logged it and then returned a bare `InvalidParameterValue`, discarding it before it ever reached the NATS hop. This is the example `mulga-e9qqx` names specifically, and it needed both fixes: the transport one alone would not have surfaced it.

Verified live on env19:

```
before: InvalidParameterValue: A value specified in a parameter is not valid, is unsupported...
after:  InvalidParameterValue: IPv6 rules are not supported
after:  InvalidParameterValue: IpPermission must specify at least one IpRange or UserIdGroupPair
```

It makes the refusal legible; it does **not** make IPv6 rules work, so `mulga-6bu36` (P0) stays open on its real ask.

Re-triage also confirmed that `mulga-m80ue`, `mulga-ue2lc`, `mulga-ps5hy`, `mulga-dye0p` and `mulga-ue2gg` are **not** fixed by this PR and are not duplicates of each other. Each discards its reason at a different call site, or is a separate defect entirely (`ue2gg` is predastore). They share this PR's remediation pattern, not its bug.

PR #1037, opened 2026-09-15
fix: answer in the envelope the caller's protocol expects, and say which service was refused

joshsiv-mulga:
Three defects at the gateway's HTTP error boundary. **None of them changes an authorisation outcome** — worth saying up front, because `mulga-dv28h` reads like an auth bypass and is not one.

## A — the XML envelope default was EC2

`xmlErrorBody` defaulted to `GenerateEC2ErrorResponse`, so any scope it did not enumerate — including every service the gateway does not serve — received the EC2 query shape `<Response><Errors>`. A REST-XML client cannot deserialize that, so the real code never surfaced.

The generic `ErrorResponse` is now the default and the query-protocol services are named explicitly.

**One deliberate carve-out.** Five `writeSigV4Error` call sites (`auth.go:42,57,59,72,80`) fire *before* `ctxService` is set at `:93`, so `svc` is `""` — the signature never parsed and no service is known. Those keep the EC2 shape. Letting them fall to the new default would turn a clean `IncompleteSignature` into a deserialization error for every EC2 client with a malformed header, which is a worse bug than the one being fixed. Consequently **no pre-existing test needed changing** for this.

Also sets `X-Amzn-Errortype` on the four JSON emitters, which JSON-protocol SDKs read to resolve the modelled exception type.

## B — the unserved-service rejection blamed the caller's credentials

`mulga-wtcab`'s own reproduction, `aws route53 list-hosted-zones`, never reached the envelope defect: `route53` is not a served scope, so the `supportedServices` gate at `auth.go:96` rejects it first. That gate logged the truthful reason and then emitted the canned credentials text, because `writeSigV4Error` had no parameter to carry a message.

The signature is never verified at that point — the gate deliberately runs before verification so a client-claimed scope cannot rubber-stamp itself — so "check your AWS credentials" was actively false. `writeSigV4Error` now takes a message and the gate names the service it refused. This is `mulga-dye0p`, and it had to land here or A could not be verified through the reproduction `wtcab` specifies.

## C — `ctxAction` was set for one protocol only

Set only from the query-arg `Action` and the admin path, so every JSON-1.1 request keyed the throttle bucket `"<account>:unknown"` — and because `ResolveLimit` was called with `"unknown"`, no per-action override could ever match. The per-action limit table was dead for those services.

Now resolved from `X-Amz-Target`, and for the path-routed rest-json services from each route's existing `lookupXAction` function. Both halves are done; `dv28h`'s acceptance criterion only named `X-Amz-Target`, which could not have covered `eks`/`bedrock*` since they send no such header.

## Verification

`make preflight` passes. Live on env19:

```
# before
route53:     SignatureDoesNotMatch: The request signature that Amazon has does not match... Check your AWS credentials
             envelope: <Response><Errors>
# after
route53:     SignatureDoesNotMatch: Service "route53" is not served by this gateway.
             envelope: <ErrorResponse><Error><Type>Sender</Type>... (botocore parses it)
autoscaling: Service "autoscaling" is not served by this gateway.
ec2:         <Response><Errors>  (unchanged — regression check)

# gateway log, previously action:"" for all three
service=eks action=ListClusters        method=GET  path=/clusters       (path-routed)
service=ecs action=ListClusters        method=POST path=/               (X-Amz-Target)
service=ecr action=DescribeRepositories method=POST path=/              (X-Amz-Target)
```

Probe TFC-0022 moved `present` to `absent`; `--cheap-only` sweep exits 0.

## Open decision for review

**TFC-0021 stays `present`, by design.** Its probe asserts the error *code* is no longer `SignatureDoesNotMatch`, and this PR keeps the code and fixes only the message. Changing it changes the HTTP status, which is a bigger behavioural call than an error-contract fix should make unilaterally. If you want the code changed too, say which — `InvalidAction` and a 404 are both defensible — and it is a small follow-up.

## Beads

- `mulga-wtcab` (P2) — fixed, narrowed. A per-service registry already existed; the defect was the `default` arm, not a missing registry.
- `mulga-dye0p` (P2, Julian) — fixed. Left open for its owner.
- `mulga-dv28h` (P2) — fixed, both halves. Not a security issue.

`mulga-zjy2g` also concerns `SignatureDoesNotMatch` from `auth.go` but is an intermittent CI-only failure with no identified code defect. It is **not** closed by this.


joshsiv-mulga:
Live-verified again on env19 against an integration build of #1037-#1041, and `mulga-dye0p` is now closed with that evidence.

One thing the full conformance run turned up that this PR does not fix, recorded so it is not mistaken for done. Probe TFC-0021 keys on the error **code**, and the code is still `SignatureDoesNotMatch`:

```
$ aws sqs list-queues
An error occurred (403) when calling the ListQueues operation: <?xml version="1.0" encoding="UTF-8"?>
<ErrorResponse><Error><Type>Sender</Type><Code>SignatureDoesNotMatch</Code>
<Message>Service &#34;sqs&#34; is not served by this gateway.</Message></Error></ErrorResponse>
```

The message is now truthful, which is what this PR set out to do. But the code still claims a signature failure that never happened, and SQS is a JSON-protocol service, so botocore cannot deserialize the XML envelope and surfaces a bare `403` — hiding the message. Filed as `mulga-5yrcl`; choosing the replacement code is a decision rather than a mechanical fix, since AWS answers a nonexistent service endpoint with a DNS failure and there is no upstream behaviour to copy.

PR #1039, opened 2026-09-15
fix(ecr): store and project repository encryption, scanning and tags

joshsiv-mulga:
Fourth cluster in the read-back conformance sweep. `CreateRepository` parsed none of `encryptionConfiguration`, `imageScanningConfiguration` or `tags` — `ecr_createrepository.go:20-21` said so outright, "accepted-and-ignored in v1" — so none were stored and none came back.

| Bead | Change |
|---|---|
| `mulga-ywvtu` | Parse, store and project `encryptionConfiguration` and `imageScanningConfiguration`; reject `KMS` |
| `mulga-j2y54` | Store repository tags; implement `TagResource` and `UntagResource` |
| `mulga-3whvb` | `PutImageScanningConfiguration` succeeds as a no-op for scanning off, still refuses scanning on |

## Why AES256 is reported but KMS is refused

This asymmetry is the substance of the change, so the evidence is worth stating.

No ECR write path sets `ServerSideEncryption` on its `PutObject` (`gateway/ecr/registry.go:432`, `:840`, `gateway/ecr/registry_image.go:94`), and there is no SSE reference anywhere under `gateway/ecr/` or `handlers/ecr/`. Read in isolation that looks like layers are stored in the clear. **They are not.** Predastore encrypts unconditionally: `Options.MasterKey` is required (`predastore.go:73-74`), `s3d` refuses to start without a key (`cmd/s3d/main.go:177-178`), the key's AEAD is handed to the blob layer (`predastore.go:339`), and every 8 KiB fragment is sealed with AES-256-GCM with its tag bound to AAD over `(objectHash, shardIndex, shardNum, fragNum)`.

So every ECR layer is encrypted at rest under a 32-byte AES-256 key, in every configuration. Reporting `AES256` states something true about the system rather than echoing a field to satisfy a planner.

`KMS` is the opposite. No customer-managed key material is plumbed anywhere, so accepting a KMS key ARN would give the caller a specific and false assurance that their key protects the data. It is rejected with `InvalidParameterValue`.

## Why scanning-on is still refused

`ScanningNotSupported` is deliberate, documented and pinned by `TestActions_ScanSurfaceUnsupported`. It returns `OperationNotSupportedException` (400) rather than `NotImplemented` (501) precisely so clients read it as a feature that is off rather than a server fault. No scanner exists — no scan, CVE, trivy or grype reference anywhere in the package.

Accepting `scanOnPush=true` and persisting it would report scanning as active while nothing scans. That is the same false-assurance trap as the ECS container fields, and worse in one respect: a security team filtering on "scan on push: enabled" would read a false all-clear.

The real defect underneath is narrower. The handler rejected the whole action **without inspecting the body**, so a caller explicitly asking for scanning **off** — the honest, already-true state — got a hard apply failure for nothing. That case now succeeds as a no-op; `true` still refuses. The other six scanning actions are untouched.

Consequently **TFC-0020 is expected to stay `present`** if its probe asserts that `scanOnPush=true` succeeds, and that is the correct outcome rather than a failure to fix.

## Projection factored first

The AWS `*ecr.Repository` was built in three places with no shared helper — `ecr_createrepository.go:78-87`, `ecr_describerepositories.go:72-79`, `ecr_deleterepository.go:82-91`. Every field this cluster adds would have been added three times, and part of why `ywvtu` forces replacement is that Create's response omitted the block too, so state never recorded a value. All three now share one `buildRepository`.

Tags follow the ECS pattern — a map on the resource's own KV record, mutated read-modify-write — rather than the EC2 tag store, which is separately S3-backed and would have meant new plumbing for no gain. `RepositoryNameFromResourceARN` was extracted from the existing authz ARN parsing so there is one parser rather than two.

## Verification

`make preflight` passes on this branch, rebased onto current `dev`.

`RepoMeta` serialises with `encoding/json` and already used `omitempty`, so existing KV records decode without migration; `meta_test.go` pins that with a zero-value record predating the new fields.

Live verification on env19 to follow once this is on `dev` alongside the other stacked branches.

One existing test narrowed rather than weakened: `TestActions_ScanSurfaceUnsupported` no longer lists `PutImageScanningConfiguration`, because that is the one action whose behaviour deliberately changed. Its other six assertions are intact.


joshsiv-mulga:

```
create-repository / describe-repositories → {"encryptionType": "AES256"}, {"scanOnPush": false}
encryptionType=KMS → InvalidParameterValueException ... no customer-managed key is used, and
                     repositories are already encrypted at rest under a server-managed AES-256 key
create-repository --tags Key=probe,Value=x Key=env,Value=prod
  → list-tags-for-resource returns both tags
create-repository --tags Key=,Value=x → InvalidParameterValueException
```

The last commit fixes something this PR originally got wrong. `createRepositoryRequest` deliberately omitted `tags`, on the reasoning that real clients tag in a follow-on `TagResource` call. AWS's `CreateRepository` does take a `tags` parameter, and live on env19 the request was accepted and the tags silently discarded — the same accepted-and-ignored shape this cluster exists to remove, and it left `mulga-j2y54` half-fixed against its own acceptance criteria. The create path now decodes and stores them.

Closes `mulga-ywvtu`, `mulga-j2y54`, `mulga-3whvb`. Conformance probes TFC-0018 and TFC-0019 moved `present` → `absent`. TFC-0020 stays `present` by design and is tracked as the missing scanner capability in `mulga-4wpwo`.

PR #1040, opened 2026-09-15
fix(iam): persist policy tags and count attachments across all principal kinds

joshsiv-mulga:
Fifth cluster in the read-back conformance sweep. Three of these are the familiar shape — the write succeeds, the projection omits a field, Terraform reads the gap as drift. One is not a read-back defect at all, and it is the reason this PR is worth reading closely.

| Bead | P | Change |
|---|---|---|
| `mulga-fvhlc` | P2 | `CreatePolicy` stores `input.Tags`; `GetPolicy`, `ListPolicies` and the create response project them |
| `mulga-9i6uh` | P3 | `AttachmentCount` counts roles and groups, not just users |
| `mulga-ryu30` | **P1** | `DeletePolicy`'s conflict guard no longer misses role and group attachments |
| `mulga-8ranb` | P3 | `ListEntitiesForPolicy` implemented |

## `mulga-ryu30` — filed during this work, and the most important item here

`buildAttachmentCounts` scanned `s.usersBucket` and nothing else. Its own doc comment was honest about it: "returns a map of policyARN -> number of users that have it attached". Roles and groups keep their attachments in separate buckets on identically shaped `AttachedPolicies` fields.

`GetPolicy` reporting 0 is the reported symptom. The consequence nobody had filed is that **`DeletePolicy` gates its `DeleteConflict` check on that same count**, so a policy attached to a role or group deleted successfully and left the principal holding a dangling ARN. Spinifex did not merely misreport the standard pre-delete safety check — it failed it. That is why this was filed at P1 rather than inheriting `9i6uh`'s P3.

## One traversal, three symptoms

`buildAttachmentCounts` was `ListEntitiesForPolicy` with the names thrown away, so both now derive from a single `buildPolicyAttachments` over users, roles and groups. Implementing the count fix alone would have meant rewriting it immediately for the reverse lookup.

**The count stays derived — deliberately.** There is no stored counter and no reverse index; the count is computed at read time from the same forward arrays `ListAttachedRolePolicies` reads, which is exactly why that call was right while the count was wrong. A maintained counter would add a drift surface where none exists, and attach and detach are CAS-guarded per principal record so they could not update a counter in the same transaction anyway.

The cost is stated plainly in a comment: this triples an already O(principals) scan on `GetPolicy`, `ListPolicies` and `DeletePolicy`. Acceptable at current scale, and visible to the next reader.

The per-key resilience of the original scan is preserved for each bucket — skip the version key, filter on the account prefix, log and continue on a missing key or unmarshal failure rather than aborting the whole scan.

## Policy tags

`CreatePolicy` built its record with `Tags: []Tag{}` hardcoded and never read `input.Tags`, and `GetPolicy`, `ListPolicies` and the create response each built an `iam.Policy` with no `Tags` field at all. The storage layer was already fine — `Policy.Tags` exists, and `TagPolicy`, `UntagPolicy` and `ListPolicyTags` all work. Both halves now match the role path.

**Why it survived:** the shared tag test created its policy *without* tags and then tagged it via `TagPolicy`, so the create path was never exercised. The new test creates with tags.

One deliberate divergence from the `CreateRole` precedent it otherwise copies: `CreatePolicy` validates tags before storing them, so an invalid tag is refused at create just as it is at tag time.

## Behaviour worth a second opinion

`ListEntitiesForPolicy` indexes the attachment map by the raw input ARN and does not first resolve the policy. Two consequences:

- An AWS-managed ARN attached to a role is returned correctly, even though no `Policy` record exists for it. That is the desirable half, and it falls out naturally rather than being special-cased.
- A genuinely nonexistent customer-managed ARN returns an empty result rather than `NoSuchEntity`, unlike `GetPolicy` and `DeletePolicy`, which still 404. If strict error parity matters here, that is the line to change.

## Not in this PR

`mulga-yxd9p` and `mulga-5eh43`, the condition-key beads originally clustered with these, are **deliberately excluded**. Their premise is that the allowlist is too narrow. Investigation found the opposite: conditions are genuinely evaluated at authorization time, unknown keys already fail closed, and `TestRegistries_EverySupportedKeyIsSuppliedByADoor` exists specifically to stop a key being allowlisted that no door supplies. Widening the allowlist without supplying a value would turn a hard `Deny` that matches everything into one that matches nothing — a silent fail-open.

They are also two different guards, not one: the `aws:RequestTag` case hits the allowlist, while the ECS trust-policy case hits a separate guard that does not consult it. Reasoning is recorded in `docs/development/bugs/iam-policy-readback.md`; the trust-policy half needs a product decision.

## Verification

`make preflight` passes on this branch, rebased onto current `dev`. No existing test was weakened or deleted. Live verification on env19 to follow once the stacked branches are on `dev`.


joshsiv-mulga:

```
create-policy --tags Key=env,Value=prod Key=owner,Value=conf
  → get-policy Policy.Tags returns both; list-policy-tags agrees
attached to one role + one group → Policy.AttachmentCount = 2
list-entities-for-policy → PolicyRoles: [conf-role], PolicyGroups: [conf-grp], PolicyUsers: []
delete-policy while attached → An error occurred (DeleteConflict) when calling the DeletePolicy
  operation: The request was rejected because it attempted to delete a resource that has
  attached subordinate entities.
after both detaches → AttachmentCount 0, delete-policy succeeds
```

The P1 (`mulga-ryu30`) is the one worth a second look on review: the old guard counted only the users bucket, so a policy attached to a role or group could be deleted out from under it. All four now share one traversal.

Closes `mulga-fvhlc`, `mulga-9i6uh`, `mulga-ryu30`, `mulga-8ranb`. Conformance probes TFC-0012, TFC-0013 and TFC-0014 moved `present` → `absent`.

joshsiv-mulga:
Rebased onto `dev` after #1037, #1038 and #1039 landed, and fixed a CI failure the integration suite caught.

`TestCrossTenant_EveryScopedActionIsAccountedFor/iam` failed:

```
iam:ListEntitiesForPolicy is neither cased nor excused — every scoped action needs
exactly one of a cross-tenant case or an entry in uncased saying why it does not need one
```

The gate was right. `ListEntitiesForPolicy` resolves `PolicyArn` through `{source: sourcePolicyARN, nameField: "PolicyArn"}` (`authz.go:105`) — the same branch as `GetPolicy` and its siblings — so the correct accounting is the `sameHelper("GetPolicy")` excuse, not a new case.

It was missing from a second gate too, which nothing failed on: `TestResourceARNs_PolicyARNIsReanchoredOntoCallerAccount`, whose comment claims "every action resolved from one is covered, so a new one cannot join by copying a neighbour". It had, so the action is now in that table as well.

`make preflight` passes on the rebased branch.

PR #1041, opened 2026-09-15
fix(acm,elbv2): serialise target group writes and fail fast on a missing tenant CA

joshsiv-mulga:
Sixth cluster in the conformance sweep. The ELBv2 half is a lost-update race, not a read-back gap, and it is the item that matters most here.

| Bead | P | Change |
|---|---|---|
| `mulga-c6khe` | **P0** | Every writer of a target group record now uses compare-and-swap |
| `mulga-wbjal` / `mulga-l6yft` | P2 | `RequestCertificate` returns a terminal, actionable error for a missing tenant CA |
| `mulga-61h3h` | P3 | `ListCertificates` returns `CertificateSummaryList` when empty |

## `mulga-c6khe` — concurrent registrations silently lost

`RegisterTargets` read the target group, appended in memory, and wrote the whole record back with an unconditional `Put`. Two concurrent registrations therefore lost one. Terraform parallelises by default, so this fired on any target group with more than one member, and **the apply exited 0 with half the fleet not behind the load balancer** — found by `terraform-aws-modules/alb 9.17.0` on a first apply against a freshly cleaned environment.

The fix uses the store's existing CAS (`kvutil.Update` with `CASConfig`), the same primitive IAM's `AttachRolePolicy` already uses for this exact shape. Contention stays per target group; no global lock was introduced, so unrelated target groups are not serialised.

**Every mutator of the record was converted, not just the one the bead names** — a CAS loop on one writer is defeated by a non-CAS writer on the same key. That is `RegisterTargets`, `DeregisterTargets`, `ModifyTargetGroup`, `ResetTargetHealthOnStartup`, the target-group branches of `AddTags` and `RemoveTags`, `ModifyTargetGroupAttributes`, and the health checker's writeback. `CreateTargetGroup` still uses a wholesale `Put`, correctly — it writes a new record with no prior read.

Retry attempts are raised to 12 from the `kvutil` default of 5, because Terraform's default parallelism is 10 writers on one record.

Two things fell out along the way: the ENI lookup in `resolveRegisteredTargetIP` now runs *before* the CAS loop so a retry does not re-issue network I/O, and a pre-existing bug where a duplicate target in a single request was appended twice is gone, because the dedup set is now updated as entries are appended.

The race tests were confirmed to fail without the fix — with the six source files stashed and the tests left in place, all three failed on multiple trials. They pass under `-race` at `-count=5`.

## `mulga-wbjal` — the retryable classification is what actually hung Terraform

Every `RequestCertificate` returned `InternalErrorException`, HTTP 500, "contact support", when the real cause was a missing tenant CA that the caller could act on themselves.

The message being discarded is half the problem. The half that hangs an apply is the classification: `InternalErrorException` is **retryable**, so the AWS provider backed off and retried rather than failing, and a single `aws_acm_certificate` sat at "Still creating..." for over nine minutes before the run was abandoned. It does not fail fast; it hangs.

It now returns ACM's own not-found code, registered at 404, which `IsTerminal` treats as terminal because 404 is in `terminalStatus`. `409` was rejected for exactly this reason — the codebase deliberately excludes it as "the caller being told to try again". No new error code was invented.

**One correction to the bead's description.** It says `privateca.go` composes an actionable message that `service_impl.go:300` then discards. In fact `LoadTenantCA`'s message is produced at daemon startup and logged there; it never reaches the handler, which only sees `s.TenantCA == nil`. The handler now composes an equivalent message from the same command-hint constant the renewal worker already uses. The daemon comment claiming `RequestCertificate` "already fails loudly with an actionable error" is now true rather than aspirational.

`mulga-l6yft` is the same defect filed a month earlier. One fix closes both.

## `mulga-61h3h`

The SDK marshaller drops a nil slice unconditionally, so an empty account returned exactly `{}`. Built with `make(..., 0, n)` now. `ListTagsForCertificate` was checked and already returns a non-nil empty slice, so there was no sibling to fix.

## Findings recorded, not fixed

- **`mulga-5rz8i` shares the lost-update shape.** `handlers/ec2/tags/service_impl.go:157` does the same read-mutate-write guarded only by a process-local mutex, so two daemons racing on one resource lose a write. It is object-store backed rather than JetStream KV, so it has no revision-checked update to reach for. Different record, different package, left alone.
- **Tag-versus-tag last-writer-wins on a target group remains.** The tag and attribute paths now CAS a single field, which stops them discarding targets, but two concurrent `AddTags` on one target group still resolve last-writer-wins on the map. Fixing that means reworking the shared taggable helpers across load balancers, listeners and rules too.

## Verification

`make preflight` passes on this branch, rebased onto current `dev`. No existing test was modified. Live verification on env19 to follow once the stacked branches are on `dev`.


joshsiv-mulga:

```
12 concurrent register-targets calls on one target group → registered=12 of 12
acm request-certificate → An error occurred (ResourceNotFoundException) when calling the
  RequestCertificate operation: no tenant CA is configured, so PRIVATE_CA issuance for
  "conf.example.com" is unavailable; create one with
  "spx admin cert create-tenant-ca --domain <domain>"
acm list-certificates (empty registry) → {"CertificateSummaryList": []}
```

The race check is 12-way rather than the 2-way the bead describes, because two calls are a weak probe for a last-writer-wins window.

Closes `mulga-c6khe`, `mulga-wbjal`, `mulga-61h3h`; `mulga-l6yft` marked a duplicate of `wbjal`. Conformance probes TFC-0009, TFC-0023 and TFC-0024 moved `present` → `absent`.

PR #1048, opened 2026-09-16
fix: decompress single-file zst/bz2/xz/gz disk images

joshsiv-mulga:
## Root cause

`ExtractDiskImageFromFile` dispatched purely on filename suffix. The `.tar.xz`/`.tar.gz`/`.tgz`/`.tar` branches worked because `tar` lists extracted members to stdout, and the shared `extractDiskImagePath` helper scanned that listing for a `.raw`/`.img` member. The plain single-file `.xz` branch reused that same helper but ran `xz -dk <path>`, which writes the decompressed file next to the source and prints nothing — so `extractDiskImagePath` parsed empty output and returned `("", nil)`, and the caller failed to stat `""`. There was no `.zst` or `.bz2` branch at all, so those fell to the final `unsupported filetype` error.

This meant every distro that publishes single-file compressed raw/qcow2 images could not be imported: Talos (`.raw.zst`, `.raw.xz`), Fedora CoreOS (`.qcow2.xz`, `.raw.xz`, `4k.raw.xz`), Flatcar (`.img.bz2`).

## Fix

- Added streaming decompression for `.zst`, `.bz2`, `.xz` and `.gz` single-file artifacts, writing the result into the caller-provided `tmpdir` and returning that path directly from the decompression step — never inferred by parsing tar chatter.
- A decompressed `.qcow2` payload (e.g. `foo.qcow2.xz`) now flows through the existing qcow2→raw `qemu-img convert` path via a new shared `finalizeDiskImage` helper, which also still runs the `file(1)` sniff gate (`validateDiskImagePath`) on every result.
- The `.tar.xz`/`.tar.gz`/`.tgz`/`.tar` branches are untouched.
- The source artifact is never modified or written next to; every decompressor streams via `io.Copy`/codec readers straight into a new file in `tmpdir`, which matters since these images are multi-GiB.

### Decompression dependency choices

- **`.zst`**: `github.com/klauspost/compress/zstd`, a pure-Go streaming decoder. It was already a transitive dependency (via `containerd`), so this promotes it to direct in `go.mod` rather than adding anything new, and avoids depending on a `zstd` binary that may not exist on a customer node.
- **`.bz2`** / **`.gz`**: standard library (`compress/bzip2`, `compress/gzip`), both streaming.
- **`.xz`**: shells out to the `xz` binary (`xz -dc` to stdout, so nothing is written beside the source and the source is untouched). No pure-Go xz decoder is already a dependency of this module, and the existing `.tar.xz` branch already requires the `xz` binary on the host (tar's `-J` invokes it internally), so this adds no new host requirement — it only makes that existing requirement explicit for the plain-`.xz` case too.

## Tests

Added a table-driven test, `TestExtractDiskImageFromFile_Compression`, in `spinifex/utils/images_test.go` covering:
- `.raw.gz`, `.raw.zst`, `.img.bz2`, plain `.raw.xz` (single-file, raw payload)
- `.qcow2.xz` (decompress then convert qcow2 → raw via `qemu-img`, using a real qcow2 fixture built with `qemu-img convert`)
- a genuinely unsupported suffix (`.raw.7z`), asserting the existing `"unsupported filetype"` error and an empty path

No fixture is downloaded — all payloads are built locally by compressing the repo's existing `tests/unit-test-disk-image.raw` fixture (gzip/zstd via the same Go libraries used in production; bzip2/xz by shelling to the system binaries, skipping the case if the tool isn't installed, matching the existing `tar`-availability guard pattern in `TestExtractDiskImageFromFile`). Every case asserts the source file's bytes are unchanged after extraction and that its directory contains no other file.

`make preflight` passes, including diff coverage (74.5%, threshold 60%).

Not covered by this PR: a live Talos `.raw.zst` import/boot check on a running environment, since the fix here is scoped to the decompression front-end. Also noticed but out of scope: `cmd/spinifex/cmd/admin.go`'s image-import command creates a real temp dir (`tmpDir`, cleaned up via `defer os.RemoveAll`) but passes the persistent per-image catalog directory (`imagePath`) to `ExtractDiskImageFromFile` instead, so extracted/converted images currently land in the catalog directory rather than being cleaned up — a separate, pre-existing bug worth its own ticket.


joshsiv-mulga:
Verified live on env13 with this branch: a Talos `.raw.zst` imported directly with no pre-decompression (`ami-a769ad10a050d6896`, `spx v1.20.0-1-gb14a67798`). The decompression front-end works.

Two follow-ups before this merges, both found by that run.

## 1. Skip the raw intermediate entirely for formats qemu can already read

`writeImageToNBD` (`spinifex/admin/image_import.go:113-125`) already writes through `qemu-img convert` and hardcodes `-f raw`:

```
qemu-img convert -f raw -O raw -n --target-is-zero -p <source> <nbd-target>
```

`qemu-img` reads qcow2 natively, so that hardcoded `-f raw` is the only reason `finalizeDiskImage` has to run `qemu-img convert -f qcow2 -O raw` into a full-size intermediate first. Passing the detected source format through instead lets a qcow2 stream straight onto the NBD export with no intermediate at all. That removes a full-size temporary for every qcow2 entry in the catalog, which today is alpine, rocky and every mulga appliance.

Probe the format explicitly with `qemu-img info --output=json` and pass the result, rather than dropping `-f` and letting qemu probe implicitly — implicit probing on an operator-supplied file is the pattern qemu warns about.

Best case this composes: a `.qcow2.xz` (what Fedora CoreOS publishes) decompresses to a ~1 GB qcow2 and that qcow2 goes straight to NBD, so an 11 GiB raw is never materialised anywhere.

## 2. The decompressors write non-sparse

Measured on the same run: the `.raw.zst` import left an **11 GiB** file in the catalog directory. Decompressing the identical artifact locally with `xz -dc | dd conv=sparse` gives **183 MiB** on disk for the same 11 GiB apparent size. The new `decompressToTemp` copies zero runs literally instead of punching holes, so it is paying 60x the space for a Talos image.

Note `writeImageToNBD` is already sparse-aware on the way out via `--target-is-zero`; only the intermediate is dense. Worth making the writer hole-punching, or using `qemu-img convert` for this hop too once point 1 is in.

## Unrelated, and please do not fix it here

`admin.go` passes `imagePath` (the persistent catalog directory) to `ExtractDiskImageFromFile` rather than the `tmpDir` it creates, so intermediates are never cleaned up. Tracked as `mulga-2t9bx`. Flagging it because this PR makes that leak reachable for artifacts that previously could not be imported at all. Its fix must NOT simply switch to `os.TempDir()`: `/tmp` on a node is tmpfs, 5.9 GB backed by 11 GB of RAM on ironbark, so that would OOM the host on any large image.

joshsiv-mulga:
Follow-up: made the decompressors sparse-aware.

**Regression, verified live on env13:** the `.raw.zst` decompression path (via `decompressToTemp`) wrote all-zero content literally instead of punching holes, so a Talos artifact that is 11 GiB apparent / 183 MiB actual with `xz -dc | dd conv=sparse` was landing as an 11 GiB file on disk — about 60x worse than the qcow2-to-raw path beside it (`qemu-img convert`, which already writes sparse: the Ubuntu intermediate on that node is 3.75 GB apparent / 2.1 GB on disk).

**Fix:** `decompressToTemp` now reads in 1 MiB blocks, `Seek`s past all-zero blocks instead of writing them, and `Truncate`s to the final size at the end so a trailing zero run is still recorded as a hole. 1 MiB amortises the read/seek syscall overhead over a multi-GiB image while remaining small enough to catch the large contiguous zero runs that dominate a sparse raw disk image; a zero run shorter than one block is written literally, which is the safe direction for a false negative to fail in. Kept fully streaming — no buffering of image content.

**Measured (synthetic, same node):** a payload that is 6 bytes of real content followed by a 2 GiB zero run: 2048.00 MiB apparent, 1.00 MiB on disk after this fix (previously would have been ~2048 MiB on disk).

**Coverage:** `zstd`, `bzip2` and `gzip` all route through `decompressToTemp`, so all three now write sparse. The plain-`.xz` path shells out to the `xz` binary with its stdout piped straight into the destination file and does **not** go through `decompressToTemp` — it's unaffected by this change and remains non-sparse. Fixing that would mean restructuring `decompressXz` to stream through the same helper (via `StdoutPipe` + explicit process lifecycle handling), which is more than this "small, self-contained" ask covers, so I left it as a follow-up rather than expanding scope. Happy to take that as a separate item if you want it closed too.

**Tests:** added `TestDecompressToTemp_WritesSparseHoles` in `spinifex/utils/images_test.go`, asserting a zero-heavy fixture's on-disk size (`stat.Blocks*512`) is under 25% of its apparent size, with content round-tripping correctly. It skips gracefully (rather than failing) if the platform doesn't expose block-count stat info or if the test filesystem doesn't appear to support sparse files. All prior assertions (source survives byte-for-byte, nothing written beside it) are unchanged and still pass.

`make preflight` passes (diff coverage 75.2%).

joshsiv-mulga:
Follow-up: plain-.xz is now sparse too, per feedback that this was in scope rather than a follow-up — .raw.xz is what Talos and Flatcar actually publish alongside .zst, so it's the path most likely hit in practice, and leaving it dense meant the same disk-size regression still fired for it.

**Fix:** `decompressXz` now hands the xz process's stdout to `decompressToTemp` via an `io.Pipe`, instead of piping straight to a file. `exec.Cmd` already runs its own copier goroutine when `Stdout` isn't an `*os.File`, and `Wait` joins it, so no extra `StdoutPipe`/manual draining is needed. A small goroutine calls `cmd.Wait()` and then closes the pipe with that result, so a read past the end of a failed decompression surfaces the real `xz` failure (stderr included) instead of a bare EOF. On the error path where `decompressToTemp` gives up early (e.g. a local write failure) before draining the pipe, the read end is closed explicitly so the copier can't stay blocked writing into it and wedge `Wait` forever.

**Measured (synthetic, same node, same 2 GiB-zero-run payload as the earlier zstd/generic measurement, this time through the real `xz` binary):** 2048.00 MiB apparent, 1.00 MiB on disk.

**Tests added** in `spinifex/utils/images_test.go`:
- `TestDecompressXz_WritesSparseHoles` — same on-disk-size assertion and graceful skip as the existing sparse test (factored into a shared `assertSparseOnDisk` helper), run through the real `xz` binary and `decompressXz`, plus asserts the source survives byte-for-byte and nothing else is written into its directory.
- `TestDecompressXz_TruncatedStreamSurfacesStderr` — truncates a compressed fixture mid-stream and asserts the returned error is not a bare `io.EOF` and surfaces xz's actual stderr ("Unexpected end of input").

Verified with `-race` as well as the normal run, given the new goroutine/pipe plumbing — clean.

`make preflight` passes (diff coverage 77.9%).

PR #1053, opened 2026-09-16
fix(ecs): enforce or refuse container runtime fields (TFC-0040)

joshsiv-mulga:
## Summary
- RegisterTaskDefinition accepted user, mountPoints, volumesFrom, readonlyRootFilesystem, privileged, pseudoTerminal, interactive, linuxParameters, systemControls and start/stopTimeout but never stored, transmitted, enforced or echoed them.
- Fields the agent can enforce (user, readonlyRootFilesystem, privileged, pseudoTerminal, interactive, systemControls, linuxParameters capabilities add/drop, startTimeout, stopTimeout) are now stored, sent to the agent, applied through containerd OCI spec options and a new graceful Stop, and echoed on describe.
- Fields the agent cannot honor (mountPoints, volumesFrom, linuxParameters devices, initProcessEnabled, sharedMemorySize, tmpfs) are refused at RegisterTaskDefinition instead of being silently dropped.
- Refusal is per value, not per field: an empty list or a false/unset flag requests nothing and is accepted, so an empty mountPoints list, as emitted by terraform-aws-modules/ecs 5.12.1, still registers.

## Test plan
- [x] Package tests for spinifex/handlers/ecs and cmd/ecs-agent all pass, including 9 new tests in taskdef_validation_test.go (echo, omission-when-unset, the mandatory empty/false-accepted case, and one rejection test per refused field) and a new bus/messages_test.go proving an older agent decoding a newer AssignContainer payload ignores unknown fields without error.
- [x] make preflight's lint and vulnerability gates pass with no findings.
- [x] make preflight from the worktree root fails only on the pre-existing, unrelated spinifex/services/predastore and spinifex/services/viperblockd integration tests, which return InsufficientStorage / storage pool at capacity, HTTP 507. Reproduced identically with this branch's changes fully stashed on the same dev tip, confirming it is an environmental/disk-pressure flake unrelated to this change, not something this PR introduced or can fix.


joshsiv-mulga:
Live verification on env19, build `f9d7c25be` (confirmed by `spx version` on all three nodes before measuring).

A real apply of the unmodified `terraform-aws-modules/ecs` module followed immediately by a plan now produces **no `container_definitions` diff at all**, and the task definition no longer forces replacement. On `6cf3a3489` five empty collections still forced one; on the build before that, twelve fields did.

```
plan rc=0   apply rc=0   drift rc=2   destroy rc=0

  # module.service.aws_ecs_service.this[0] will be updated in-place
  ~ resource "aws_ecs_service" "this" {
      + availability_zone_rebalancing      = "DISABLED"
        # (14 unchanged attributes hidden)
    }
Plan: 0 to add, 1 to change, 0 to destroy.
```

Reproduced on a second full cycle with byte-identical drift. The one remaining attribute is unrelated to this PR: `DescribeServices` does not return `availabilityZoneRebalancing`, and the field is absent from `aws-sdk-go v1.55.8` which `go.mod` pins, so `ecs.Service` has nowhere to carry it. Filed separately as `mulga-t239t`.

`probe_TFC_0027`, `probe_TFC_0028` and `probe_TFC_0040` all read `absent` against the live stack. Three full apply/destroy cycles left zero orphaned ENIs.

Not covered by this run: agent-side enforcement (`readonlyRootFilesystem` actually read-only inside a running container, the process actually running as the requested user). The ecs-agent ships in the ecs-node AMI, not via `cluster-deploy`, so that needs `make build-ecs-node-image` plus a reimport — as does the `heldStdin` fix in `6cf3a3489`.

PR #1054, opened 2026-09-16
fix(ecs-agent): allocate a console for a pseudoTerminal container

joshsiv-mulga:
## Summary

`pseudoTerminal: true` on a container definition registered fine, was echoed back by `DescribeTaskDefinition`, and then made every task using it unstartable. The task went straight to STOPPED with no exit code:

```
container start failed: create task <id>-probe: failed to create shim task:
OCI runtime create failed: runc create failed: cannot allocate tty if runc
will detach without setting console socket
```

`Run` added `oci.WithTTY`, which puts `Terminal: true` in the OCI spec, while the task was created with a cio creator that never asked for a terminal — `cio.NullIO` for a non-interactive container, a plain stream creator for an interactive one. `cio.NewCreator` builds its FIFO set from `Streams.Terminal`, so with that flag unset the shim creates no console socket and runc refuses the pair outright.

Isolated on env19 with two separate task definitions: `interactive: true` alone runs and exits 0, `pseudoTerminal: true` alone fails every time, and the two together fail the same way. Found by the agent-side verification sweep in #1053, which was only possible once the ecs-node AMI was rebuilt.

## Changes

- `ioStreamOpts(stdin io.Reader, terminal bool) []cio.Opt` builds the stream options in one place. A terminal container gets `cio.WithTerminal` and a nil stderr, because a terminal carries no separate stderr stream and the shim merges it into the console.
- `containerIOCreator` takes the terminal flag alongside interactive. `cio.NullIO` is now used only when the container asks for neither stdin nor a terminal, and a terminal container that is not interactive gets no stdin stream at all.
- `Run` lifts the pseudoTerminal test into a `terminal` variable and passes it to the creator.

The alternative was to refuse the field at `RegisterTaskDefinition` the way `mountPoints` and `volumesFrom` are. Honouring it is the better call: the runtime already carries the field end to end, the missing piece is one cio option rather than a subsystem, AWS allocates a TTY rather than rejecting it, and refusing would walk back a field the API currently reports as stored.

## Test plan

- [x] Unit: the terminal path sets `Terminal` and clears stderr, the non-terminal path keeps stderr and stdin, and a terminal-without-interactive container gets a creator while registering no stdin to release. The existing `heldStdin` tests still pass, so the interactive path is unchanged.
- [ ] Live on env19: rebuild and reimport the ecs-node AMI, then run one task with `pseudoTerminal: true` and one with `interactive: true`. Both must reach STOPPED with an exit code rather than a create failure. Result will be added as a comment.

Closes `mulga-2lk0y`. Conformance gap TFC-0044.


joshsiv-mulga:
Verified live on env19.

The ecs-node AMI was rebuilt from this branch and imported as `ami-fccb2cf3d4b209393`; `DescribeInstances` confirms the container instance ran that AMI before any measurement below.

Same two task definitions that isolated the bug, run again on the fixed agent:

```
agentverify-int-only    STOPPED  0  container probe exited (0)
agentverify-tty-only    STOPPED  0  container probe exited (0)
```

Before the fix the tty case was `STOPPED  None  container start failed: ... cannot allocate tty if runc will detach without setting console socket`. It now starts, runs its command and reports an exit code, and `interactive: true` still exits 0 on its own, so the existing path is unchanged.

`probe_TFC_0044` moves from `present` to `absent`.

PR #1055, opened 2026-09-16
feat: report availabilityZoneRebalancing on ECS services

joshsiv-mulga:
## Summary

`DescribeServices` returned no `availabilityZoneRebalancing` field. The terraform provider treats the attribute as computed, so it read empty and planned an in-place update to `DISABLED` on every plan — including one taken immediately after a clean apply:

```
  # module.service.aws_ecs_service.this[0] will be updated in-place
  ~ resource "aws_ecs_service" "this" {
      + availability_zone_rebalancing      = "DISABLED"
    }
Plan: 0 to add, 1 to change, 0 to destroy.
```

This is the only diff left on the `ecs` conformance module, so it is the last thing standing between that module and a `No changes` plan.

## Why the field cannot come from the SDK

`go.mod` pins `github.com/aws/aws-sdk-go v1.55.8`, the last v1 release. AWS added `availabilityZoneRebalancing` after v1 stopped taking model updates, so no v1 version carries it (`go list -m -versions` tops out at v1.55.8) and `ecs.Service` has no such member. Two alternatives were tried and rejected:

- **Embedding `*ecs.Service` in a wrapper carrying the extra field.** `jsonutil.BuildJSON` nests an embedded struct rather than flattening it, so the service arrives under a `Service` key. Confirmed by experiment, not assumed.
- **Mirroring `ecs.Service` as a local type.** Forty-odd fields duplicated to carry one string, and every future SDK change has to be mirrored too.

## Changes

- `checkAZRebalancing` parses the raw `CreateService` / `UpdateService` body and refuses anything but `DISABLED` with `InvalidParameterException`. Nothing here rebalances tasks across zones, so accepting `ENABLED` would report a service as rebalancing when nothing does. A body that does not parse defers to the SDK shapes, which produce the real error.
- `addSDKGapFields` decodes the marshalled response, adds the field to each service object, and re-encodes. It runs from `WriteJSONResponse`, so every service-carrying action is covered from one place and a response carrying no service is returned untouched. The decode uses `json.Decoder.UseNumber`, because a `float64` round trip would re-render the epoch-second times `jsonutil` emitted in exponent form. A value already present is left alone, so this never overwrites a real answer once the field has a source.

## Test plan

- [x] Unit: the refusal accepts absent/`DISABLED` and refuses `ENABLED` with the right code; the injection covers `DescribeServices`, `CreateService`, `UpdateService` and `DeleteService`, leaves other responses byte for byte, preserves the time encoding, and keeps an existing value.
- [ ] Live on env19: deploy and re-run the `ecs` conformance module. The drift plan must come back `No changes`, which would be a first for that module. Result will be added as a comment.

Closes `mulga-t239t`. Conformance gap TFC-0042.


joshsiv-mulga:
## Live verification — env19

Deployed this branch to env19 (3 nodes, rolling, `ok=50 changed=8 failed=0` on each), then re-ran the `ecs` conformance module. Existing state was destroyed first so the apply measured this build rather than converging on top of an older one.

```
{
  "module": "ecs",
  "plan_rc": 0,
  "apply_rc": 0,
  "drift_rc": 0,
  "destroy_rc": 0
}
```

`drift_rc: 0` is the acceptance criterion, and the drift plan reads:

```
No changes. Your infrastructure matches the configuration.
```

This is the first time the `ecs` module has produced a `No changes` plan. Reproduced on a second run (`--keep`), which came back `drift_rc: 0` as well.

The conformance probe moved with it:

```
before:  present  P2  TFC-0042  DescribeServices omits availabilityZoneRebalancing
after:   absent   P2  TFC-0042  DescribeServices omits availabilityZoneRebalancing
```

The refusal path was checked against the live service too:

```
$ aws ecs describe-services --cluster tfc2-ecs --services tfc2-svc \
    --query 'services[0].[serviceName,availabilityZoneRebalancing,status]'
tfc2-svc	DISABLED	ACTIVE

$ aws ecs update-service --cluster tfc2-ecs --service tfc2-svc \
    --availability-zone-rebalancing ENABLED
An error occurred (InvalidParameterException) when calling the UpdateService operation:
availabilityZoneRebalancing ENABLED cannot be applied (no cross-zone rebalancer)

$ aws ecs update-service --cluster tfc2-ecs --service tfc2-svc \
    --availability-zone-rebalancing DISABLED --query 'service.[serviceName,availabilityZoneRebalancing]'
tfc2-svc	DISABLED
```

So `DISABLED` round-trips through `UpdateService` and `DescribeServices`, and `ENABLED` is refused rather than silently accepted.

PR #1056, opened 2026-09-16
fix: carry AWS device name in virtio-blk serial for guest udev remap

joshsiv-mulga:
## Summary
- The aws-ebs-csi-driver requests an AWS-style device name (e.g. `/dev/xvdaa`) and Spinifex echoes it in `DescribeVolumes`, but the guest kernel names virtio-blk devices sequentially (`/dev/vdb`), so `MountVolume.MountDevice` fails with `NotFound: Failed to find device path`.
- Mirrors AWS Nitro's own mechanism: the requested device name is normalized (strip `/dev/`) and written into the virtio-blk `serial=` property, validated against the 20-byte cap and the driver's fatal `vol[a-z0-9]+` serial check, so a guest-side udev rule can rebuild the `/dev/xvd*` symlink by string substitution.
- RDS and Bedrock explicitly opt into the pre-existing volume-ID-derived serial via a new `serial` parameter threaded through `AttachVolume`, since RDS's device is a fixed per-slot convention and its format-authorization check (`docs/development/archive/bugs/rds-data-volume-format-authorization.md`) depends on the volume-ID-derived value. Nothing about RDS's guest device-identity mechanism or security posture changes.
- The udev rule (belongs with future demo/provider Talos machine-config material):
  ```
  SUBSYSTEM=="block", KERNEL=="vd*", ATTR{serial}=="xvd*", SYMLINK+="$attr{serial}"
  ```
- What is lost: nothing for RDS. For generic/CSI volumes, the serial no longer carries the volume ID, so the driver's advisory-only upstream cross-check (`verifyVolumeSerialMatch`, a log line not a mount failure, confirmed against upstream `master`) has nothing to compare against going forward.

## Live verification (env13, 2026-09-16)
Deployed this branch to the single-node `ironbark` env (no submodule pointer bump), launched a Talos v1.11.0 control-plane node with the udev rule above, installed Cilium + `aws-ebs-csi-driver` pointed at Spinifex's gateway with a scoped (non-admin) IAM role, and proved the round trip:
- `DescribeVolumes` showed the exact failure precondition: `vol-1891d9922b1d53b76 /dev/xvdaa attached`.
- `MountVolume.MountDevice` succeeded; in-guest `xvdaa -> vdb` symlink was present.
- Sentinel written by pod 1, pod deleted, `AttachVolume.Attach succeeded` fired for a fresh pod against the same PVC (`vdb`'s device-node timestamp advanced, proving a real reattach not a stale mount), and the sentinel read back identical.
- Full details, including two unrelated IMDS/cloud-provider fidelity gaps hit and worked around (not touching Spinifex code), are in `docs/development/archive/bugs/virtio-ebs-device-serial.md` in the mulga superrepo.
- Not exercised: reschedule onto a different node (single-node env).

## Test plan
- [x] `make preflight` green
- [x] `go test ./...` green in touched packages (`vm`, `daemon`, `handlers/rds`, `handlers/bedrock`), including new `TestAttachmentSerial_NormalizesAndValidates`
- [x] Invariant tests (`vm/lifecycle_invariants_test.go`, `vm/gc_invariants_test.go`, `daemon/vm_adapters_invariants_test.go`) pass unchanged
- [x] Live sentinel round-trip on env13 (see above)


joshsiv-mulga:
Closing unmerged. The change works and is live-verified, but it is the wrong shape: it asks every guest OS to carry a Spinifex-specific udev rule before the EBS CSI driver can mount, and it moves the virtio serial away from the volume ID, which is what AWS's own NVMe serial carries. It also forced caller-conditional serial behaviour so RDS and Bedrock could keep their existing semantics, which means the serial now means different things depending on who attached the volume.

Real EC2 carries device identity in the virtual hardware rather than asking the guest to adapt. On Nitro the NVMe controller reports model Amazon Elastic Block Store with the volume ID as its serial, and stock udev builds the by-id path the driver looks for with no guest tooling at all. On pre-Nitro Xen the guest receives the requested device name from the hypervisor directly. Spinifex offers neither, and that is the actual deficiency worth fixing.

Reopening mulga-jr07b and spiking whether the AWS-faithful path is reachable, along with what else in the stack this same gap affects.

PR #1057, opened 2026-09-16
fix(ec2): render image CreationDate as UTC

joshsiv-mulga:
## Summary

An AMI imported on a node whose timezone is not UTC gets a `CreationDate` that is local wall-clock time carrying a `Z` suffix, so on a +1000 node it reads ten hours in the future:

```
spinifex-ecs-node          ami-f931e4c210c91f28e  2026-09-16T21:42:24.895Z
ami-alpine-3.21.7-x86_64   ami-a07a7ed864cba8787  2026-09-16T10:56:37.243Z
```

The node's clock read `2026-09-16T11:42:52Z` at the moment of that import, so an 11:42Z event was stamped 21:42Z. The Ubuntu AMI above it, registered through a different path, is correct UTC — the two disagreed by exactly the local offset.

The layout `"2006-01-02T15:04:05.000Z"` treats the trailing `Z` as a **literal character**, not a zone directive. `Format` therefore printed whatever zone the `time.Time` carried and then asserted UTC regardless.

## Changes

- `describeImages` converts before formatting. This is the single authoritative render site, so it also corrects timestamps already stored, because the stored `time.Time` still carries its original zone.
- The four sites that stamp a new image — `CreateImageFromInstance`, `CopyImage`, `RegisterImage` and `spx admin images import` — store `time.Now().UTC()`, so the value is right on both sides of the round trip rather than only on the way out.

Two nearby sites were checked and needed no change: `images_remove.go` only copies the `time.Time` without formatting it, and the two `spx admin` display paths already call `.UTC()` before formatting.

## Why it matters

Terraform's `most_recent = true` sorts on this field, so an AMI imported on a +1000 node outranked a genuinely newer one imported on a UTC node for ten hours. Any comparison against a real clock was wrong by the offset.

## Test plan

- [x] `TestDescribeImages_CreationDateRenderedAsUTC` builds a record whose `CreationDate` sits in `time.FixedZone("test", 10*60*60)` at wall clock `21:42:24.895` and asserts the exact rendered string `2026-09-16T11:42:24.895Z`. A fixed zone rather than `time.Local` keeps it deterministic on a UTC runner. Confirmed to fail (reporting `21:42:24.895Z`) with the `.UTC()` removed.
- [ ] Live on env19: re-import an AMI and confirm `DescribeImages` reports the UTC instant, and that `probe_TFC_0045` moves from `present` to `absent`. Result will be added as a comment.

Closes `mulga-il7w1`. Conformance gap TFC-0045.


joshsiv-mulga:
## Live verification — env19

Deployed to env19 (3 nodes, rolling, `ok=50 changed=8 failed=0` on each). The build was a scratch merge of this branch with #1055 and #1058, which touch disjoint packages; this change is the only one in `ec2/image`.

The fix corrects records already stored, which this shows directly — the two `spinifex-ecs-node` AMIs below were imported *before* the deploy, under the old code, and now render correctly without being re-imported:

```
$ aws ec2 describe-images --query 'Images[].[Name,ImageId,CreationDate]' --output text
spinifex-eks-node          ami-16e62bfdfaa7cef6b  2026-09-16T06:08:56.938Z
ami-alpine-3.21.7-x86_64   ami-5d48ef63a080da44a  2026-09-16T00:56:24.715Z
ami-ubuntu-26.04-x86_64    ami-6827189291e989525  2026-09-16T00:46:45.023Z
ami-ubuntu-26.04-x86_64    ami-9e64e61b06d810a08  2026-09-16T00:55:35.000Z
ami-alpine-3.21.7-x86_64   ami-a07a7ed864cba8787  2026-09-16T00:56:37.243Z
spinifex-ecs-node          ami-f931e4c210c91f28e  2026-09-16T11:42:24.895Z
spinifex-ecs-node          ami-fccb2cf3d4b209393  2026-09-16T12:19:54.713Z

$ ssh casuarina date -u +%Y-%m-%dT%H:%M:%SZ
2026-09-16T13:33:20Z
```

`ami-f931e4c210c91f28e` read `2026-09-16T21:42:24.895Z` before this change and reads `11:42:24.895Z` now, which is the UTC instant of that import. Every timestamp moved back by the node's +1000 offset, and every one is now in the past relative to the node clock — before, three of them sat hours in the future.

The probe moved with it:

```
before:  present  P3  TFC-0045  An imported image's CreationDate carries local time under a Z suffix
after:   absent   P3  TFC-0045  An imported image's CreationDate carries local time under a Z suffix
```

PR #1058, opened 2026-09-16
fix(ecs): retry an owed awsvpc task ENI release

joshsiv-mulga:
## Summary

Releasing an awsvpc task ENI was best effort. `reclaimTaskENI` logged the failure and returned, and nothing ever tried again:

```go
if err := s.eni.Release(ctx, accountID, task); err != nil {
	slog.ErrorContext(ctx, "ECS: task ENI release failed",
		"task", task.TaskID, "eni", task.ENIID, "err", err)
}
```

Its own comment said the ENI fields "stay on the record for a later retry", but no later retry existed. The stopped-task sweep pruned the record an hour after the task stopped without looking at those fields, so the only pointer to the leaked interface was deleted and the ENI was orphaned permanently.

An orphaned task ENI still references its security group, so `DeleteSecurityGroup` fails with `DependencyViolation` and no awsvpc stack can be destroyed without deleting the ENIs by hand. Terraform retries the group delete for about fifteen minutes before failing.

The three ordinary teardown paths — graceful stop, `forceStopTask`, and the agent-reported STOPPED transition — all release correctly on a healthy cluster, confirmed on env19. The defect is entirely in what happens when that one call fails.

## Root cause, in four parts

1. **Success was not recorded.** `reclaimTaskENI` never cleared `ENIID`/`ENIAttachmentID`, so a non-empty `ENIID` on a stopped record meant "released fine" and "still owed" equally. Nothing can retry from a signal that does not distinguish the two.
2. **Nothing retried.** `sweepStoppedTasks` is the only pass that walks stopped records, and it only pruned.
3. **The pointer was destroyed.** `sweepStoppedBucket` deleted any record stopped longer than `stoppedTaskRetention` regardless of whether its ENI was ever released.
4. **The placement rollback leaked silently.** In `provisionTaskENI` an attach failure called `reclaimTaskENI`, then cleared the ENI fields unconditionally, and `RunTask`'s `continue` meant the record was never persisted. A release that failed there leaked an ENI with no record of it anywhere.

## Changes

A non-empty `ENIID` on a stopped record now means exactly one thing — this ENI is still owed — and the existing sweep pays the debt off.

- `reclaimTaskENI` clears the identity when `Release` returns nil and leaves it when it does not. `forceStopTask` moves its reclaim above its `putJSON` so the cleared fields land in the single write it already does, matching what `releaseTaskPublicIP` does directly above it; the other two call sites already persist afterwards.
- `TaskRecord` carries `ENIReleaseAttempts` and `ENIReleaseNextTry`, both omitted when zero, so a record with no ENI trouble serialises exactly as before.
- `sweepStoppedBucket` retries an owed release, clearing and persisting on success and backing off on failure. Backoff doubles from 30s to a 15m ceiling. At most twenty are attempted per pass, because `Release` is a 30s-budget NATS round trip and a bucket full of them would stall the sweep for the whole fleet; a record deferred by that cap asks for the pass back at `sweepInterval` rather than contributing no deadline.
- A record that still owes an ENI is never pruned, however far past retention it sits. It is retired when the release succeeds. These are bounded by the number of tasks whose release actually failed, which on a healthy cluster is zero.
- The attach-failure rollback clears the ENI fields only when the release succeeded. When it did not, the record is persisted as STOPPED with the fields intact so the sweep owns it.

### Rejected: sweep EC2 for orphaned ENIs

Listing every ENI described `ecs-awsvpc-task` that is `available` and deleting the ones no live task claims would also catch leaks from a scheduler that died mid-placement. It is rejected because the window between `Allocate` and `Attach` is exactly an ENI that is `available` and claimed by no persisted task, so that reconciler would race live placements and delete ENIs out from under them. A record-driven retry has no such window.

## Behaviour change worth flagging

A caller that gets a `RESOURCE:eni` failure from `RunTask` will now see a STOPPED task in `ListTasks`/`DescribeTasks` that AWS would not show, in the specific case where the ENI release also failed. This is deliberate — a visible stopped task carrying a `StoppedReason` beats a silently leaked ENI. It was checked against the handlers that react to stopped tasks: `recordDeploymentFailure` is reached only from the agent/bus-reported STOPPED transition, not from this path, so the deployment circuit breaker is not fed; service task counting filters on `LastStatus`/`DesiredStatus` so the record cannot be counted as running or pending; the reaper skips records already STOPPED; and capacity is still released exactly once, by the existing `rollback` call.

## Test plan

- [x] Unit: a release that succeeds clears the identity and one that fails leaves it; the attach rollback persists a STOPPED record only when the release failed; the sweep retries an owed ENI, backs off on failure, skips one that is not yet due, honours the per-pass cap, and asks for `sweepInterval` when the cap defers a due retry; the backoff doubles and caps.
- [ ] Live on env19: apply the `ecs` conformance module and confirm it still destroys cleanly, then force a release failure and confirm the record survives the sweep and the ENI is deleted on a later pass. Result will be added as a comment.

Closes `mulga-1ysjx`. Conformance gap TFC-0041.

Note on the probe: `probe_TFC_0041` looks for an orphaned `ecs-awsvpc-task` ENI, and every ordinary teardown path releases correctly, so it reads `absent` on a healthy cluster and cannot by itself gate this fix. The live test has to force the release to fail.


joshsiv-mulga:
## Live verification — env19

Deployed to env19 (3 nodes, rolling, `ok=50 changed=8 failed=0` on each, predastore quorum ok). The build was a scratch merge of this branch with #1055 and #1057, which touch disjoint packages (`gateway/ecs` and `ec2/image`); this change is the only one in `handlers/ecs`.

The rolling deploy was itself the injection. It restarted `spinifex.target` on all three nodes with the `ecs` conformance stack up and an awsvpc service running, which is exactly the condition that produced the original leak — the gap was found when a `cluster-deploy` killed a module's tasks and left two orphaned `ecs-awsvpc-task` ENIs behind.

After the deploy, with 213 task records in the cluster and the service still cycling tasks:

```
$ aws ec2 describe-network-interfaces --filters Name=status,Values=available \
    --query 'length(NetworkInterfaces)'
0
```

No interface was left `available`. Then the stack was destroyed:

```
{
  "module": "ecs",
  "destroy_rc": 0
}
```

`destroy_rc: 0` — the security group deleted without a `DependencyViolation`, and no task ENI survives:

```
$ aws ec2 describe-network-interfaces \
    --query 'NetworkInterfaces[?contains(Description, `ecs-awsvpc`)].[NetworkInterfaceId,Status]'
(empty)
```

`probe_TFC_0041` reads `absent`, as it did before this change. That is expected and is why it cannot gate this fix: the probe looks for an already-orphaned interface, and every ordinary teardown path releases correctly. What the live run establishes is that the happy path is unregressed under the exact scenario that originally leaked. The retry path itself — release fails, record survives the sweep, a later pass succeeds and clears it — is covered by the unit tests, since forcing a NATS-level `DeleteNetworkInterface` failure on a live cluster is not reachable from the API surface.

PR #1059, opened 2026-09-16
fix(ecs): keep a stopped task's ENI attachment after its release

joshsiv-mulga:
## Summary

`DescribeTasks` returns an empty `attachments` array for every STOPPED awsvpc task. On env19 against `dev@0fd6848fb`, 134 of the 135 tasks in the `tfc2-ecs` cluster reported `attachments: []`, and only the single PENDING task carried an `ElasticNetworkInterface` block.

On AWS a stopped task keeps its ENI attachment for the stopped-task retention window, with `status: DELETED` and the `networkInterfaceId` still readable. It is part of the forensic record of a task that has stopped.

## Root cause

This is a regression from #1058, caught by the post-merge conformance sweep.

`taskToAWS` gates the whole attachment block on `r.ENIID != ""`. #1058 made `reclaimTaskENI` clear `ENIID` and `ENIAttachmentID` on a successful release, so that a non-empty `ENIID` would mean "this ENI is still owed" and the stopped-task sweep could retry from it. The signal was right; carrying it in the *emptiness* of `ENIID` was not, because the same field is the projection's only source.

## Changes

The owed/released state moves into a field of its own.

- `TaskRecord` gains `ENIReleased bool`, omitted when false.
- `reclaimTaskENI` sets it on success and leaves `ENIID`/`ENIAttachmentID` in place. Its guard gains `|| task.ENIReleased`, so a second call on an already-released record is a no-op rather than a repeat delete.
- The sweep's owed test becomes `ENIID != "" && !ENIReleased`, and its refusal to prune keys on the same test, so a released record ages out on retention exactly as it did before #1058. `retryTaskENIRelease` reads the flag rather than an emptied `ENIID`.
- The placement rollback tests `rec.ENIReleased` for "release succeeded". Behaviour is otherwise unchanged.
- The attachment reports `Status: DELETED` once released, rather than echoing the task's own `LastStatus`.

`ENIID` once again means what it says — this task had an ENI — so the gate on it is now correct as written.

## Not affected

`assignTaskPublicIP` is guarded by `msg.LastStatus == TaskStatusRunning && prev != TaskStatusRunning` at its only call site, so it fires solely on the transition into RUNNING. A released stopped task cannot re-enter the auto-EIP path; the protection is that guard, not the emptiness of `ENIID`. `releaseTaskPublicIP` keys off `ENIEIPAllocationID` and is likewise unaffected.

## Test plan

- [x] `TestTaskToAWS_StoppedAwsvpc_ReleasedENI_ReportsDeletedAttachment` pins the regression: a STOPPED awsvpc record with `ENIReleased` still projects the attachment with `networkInterfaceId`, `privateIPv4Address` and `Status` `DELETED`.
- [x] `TestSweepStoppedBucket_PrunesReleased_KeepsOwed` pins both halves of the pruning rule in one test.
- [x] The tests that asserted on a cleared `ENIID` now assert on `ENIReleased`, and the reclaim test also checks that a second call does not re-release.
- [ ] Live on env19: deploy and confirm `DescribeTasks` on a stopped awsvpc task carries the attachment, and that `probe_TFC_0041` moves from `skipped` to a real verdict. Result will be added as a comment.

Closes `mulga-5bnx3`.

## Note on how this was found

`probe_TFC_0041` follows a stopped task to its ENI to decide whether one leaked. With no attachment to follow it reported `skipped` on every run, which is how the regression surfaced — the probe for the P1 that #1058 fixed had been silently disabled by #1058 itself.


joshsiv-mulga:
## Live verification — env19

Deployed this branch to env19 (3 nodes, rolling, `ok=50 changed=8 failed=0` on each). `spx version` on the init node reads `v1.20.0-20-g1501b54af`, so the running build is this commit.

Before the deploy, on `dev@0fd6848fb`, no stopped task in the `tfc2-ecs` cluster carried an attachment:

```
total arns 187
stopped 186 non-stopped 1
stopped with attachments: 0
```

After the deploy, the service kept cycling tasks, and every task that stopped under the new build carries the block:

```
total 221  stopped 221  stopped-with-attachment 32
attachment statuses seen: {'DELETED'}
```

A stopped task now reads as AWS returns it:

```json
{
  "id": "eni-attach-e2b71d4e7f267ed06",
  "type": "ElasticNetworkInterface",
  "status": "DELETED",
  "details": [
    {"name": "networkInterfaceId",  "value": "eni-46cc6522f3225117f"},
    {"name": "privateIPv4Address",  "value": "10.232.1.5"},
    {"name": "macAddress",          "value": "02:d7:8c:b8:9e:c0"},
    {"name": "subnetId",            "value": "subnet-4bf16abf26afcb43f"}
  ]
}
```

The 189 records that predate the deploy still read `attachments: null`, and always will. They were written by #1058's code, which cleared `ENIID` on disk on release, so the identity this change projects from is gone from those records. This fix stops the loss; it cannot undo it. That is the whole of the difference between 32 and 221 above — not a partial fix.

The release path itself is unregressed:

```
$ aws ec2 describe-network-interfaces --filters Name=status,Values=available \
    --query 'length(NetworkInterfaces)'
0

$ aws ec2 describe-network-interfaces \
    --query 'NetworkInterfaces[?contains(Description, `ecs-awsvpc`)].[NetworkInterfaceId,Status]'
(empty)
```

The probe that this regression had disabled is live again:

```
before:  skipped  P1  TFC-0041
after:   absent   P1  TFC-0041
```

Worth repeating from the run record for #1058: `absent` on TFC-0041 is evidence of a healthy cluster, not evidence the retry works. The result that matters here is `skipped` → a real verdict, which is the probe being able to run at all.

PR #1060, opened 2026-09-16
fix(vpc): accept security-group rules carrying an IPv6 range

joshsiv-mulga:
## Summary

`AuthorizeSecurityGroupIngress`/`Egress` rejected any `IpPermission` carrying an `Ipv6Ranges` entry with `InvalidParameterValue: IPv6 rules are not supported`. Terraform's `aws_security_group` resource sends egress as a single permission set, so a module that requests `["0.0.0.0/0"]` and `["::/0"]` together — which is the default in `terraform-aws-modules/vpc` and `terraform-aws-modules/security-group` — got the whole call refused and the group ended with **no egress at all**, not merely no IPv6 egress.

IPv6 ranges are now parsed, validated, stored, and projected back on `DescribeSecurityGroups` and `DescribeSecurityGroupRules`. They stop at the network layer: `toPolicyRules` and `sgRulesToPolicyRules` drop any rule with no IPv4 source, because the ACL builder is IPv4-only and a rule reaching it with an empty CIDR and an empty SourceSG is an unspecified ACL, not a narrower one. An IPv6 rule is therefore inert — which is exactly what it is on AWS in a VPC with no IPv6 CIDR.

## Changes

- `SGRule` gains `CidrIpv6`; validation accepts it as a third source alongside `CidrIp` and `SourceSG`, and rejects an IPv4 CIDR in the v6 field, a non-canonical CIDR, and a rule with no source at all.
- `sgRuleKey` includes the address family, so a v4 and a v6 rule on the same port are distinct rather than colliding as duplicates.
- `ipPermissionsToSGRules` and `collectDescriptionTargets` parse `Ipv6Ranges` in both authorize and revoke modes.
- `sgRulesToIpPermissions` and `sgRuleToSecurityGroupRule` project the v6 range back.
- `toPolicyRules` (subscribers) and `sgRulesToPolicyRules` (reconcile) skip rules with no IPv4 source.

## Behaviour change worth review

Revoking an IPv6 rule that does not exist now returns `InvalidPermission.NotFound` rather than succeeding as a no-op. The no-op existed only because authorize refused to store the rule, so a caller's state could never match; with authorize working there is nothing left to paper over. The default SG egress is IPv4-only, matching AWS for a VPC with no IPv6 CIDR, so the provider will not blind-revoke a `::/0` rule it never created.

## Testing

`make preflight` green, all 49 test packages pass. New and rewritten unit tests cover: IPv6-only authorize and revoke, a mixed v4/v6 permission in both modes, address family as part of the rule key, the cross-family validation rejections, the describe round trip, description updates on a v6 rule, and both policy-rule converters dropping v6.

Live verification on env19 is with the conformance harness: the `vpc` module applies with `egress_ipv6_cidr_blocks` at its default and the group ends with both egress rules rather than none.


joshsiv-mulga:
Live-verified on env19 (casuarina .19 init, bottlebrush .20, ironbark .21) after a rolling `cluster-deploy` of this branch.

**A single permission carrying both address families is accepted and both come back.** This is the shape the Terraform provider sends:

```
IpProtocol "-1"
  IpRanges    0.0.0.0/0, 10.229.99.0/24
  Ipv6Ranges  2001:db8::/32
```

**The module that found the gap now applies clean.** `terraform-aws-modules/security-group` 5.3.1 with `egress_rules = ["all-all"]` and `egress_ipv6_cidr_blocks` left at its `["::/0"]` default: `plan_rc 0, apply_rc 0, drift_rc 0`, and the group ends with the allow-all egress rule carrying `0.0.0.0/0` and `::/0` plus their descriptions. Before the fix the whole authorize was refused, and because the provider had already revoked the default rule the group was left with no egress at all.

**The IPv6 half stops at the policy layer rather than reaching the ACL builder as an unspecified rule.** OVN on the live group:

```
from-lport  1000 (inport == @sg_... && ip4) allow-related
from-lport   800 (inport == @sg_...) drop
  to-lport  1000 (outport == @sg_... && ip4 && tcp.dst == 22 && ip4.src == 192.168.1.33/32) allow-related
  to-lport   900 (outport == @sg_...) drop
```

One `ip4` egress ACL from the IPv4 half, nothing from the IPv6 half, ingress still scoped to the operator CIDR.

The conformance probe was hardened before this was measured — it previously only checked the authorize call did not return `InvalidParameterValue`, so a call that succeeded and stored nothing would have read as fixed. It now requires `::/0` to come back from `DescribeSecurityGroups`. It reads `absent`, and the full sweep is `absent=33 present=9 skipped=4 error=0`.

PR #1062, opened 2026-09-16
fix(ecs): honour ListTasks filters and count instance tasks by status

joshsiv-mulga:
Closes part of `mulga-gm5l3` (Ben's bead — left open, annotated, for the two items below that are not in this PR). Conformance gap TFC-0043.

## What was wrong

`ListTasks` read only `input.Cluster` and returned every task record under the cluster prefix. `DesiredStatus`, `Family`, `ServiceName`, `StartedBy` and `ContainerInstance` were accepted and discarded. Measured on a live cluster at 1 RUNNING against 104 STOPPED: both status filters returned the same 105 ARNs.

`RunningTasksCount` on a container instance came from `len(InstanceRecord.PlacedTasks)`. `PlacedTasks` is the set of tasks still owing capacity to the instance, not a status tally, so a PENDING task counted as running and a task whose reservation release did not complete counted as running forever. `PendingTasksCount` was never set.

## What changed

- A `taskFilter` built from `ListTasksInput` and applied per record. An omitted `desiredStatus` means RUNNING, as it does on AWS. Filters combine with AND.
- `recordTaskState` moves a self-exited task's `DesiredStatus` to STOPPED alongside `LastStatus`. Without this the filter would be honoured and still give the wrong answer, because a task that died on its own would keep answering the RUNNING filter.
- `instanceTaskCounts` tallies RUNNING and PENDING per instance from the task records — the same treatment `applyClusterCounts` already gives the cluster-level counts — and `instanceToAWS` sets both counts from it.
- `DeregisterContainerInstance` passes zero counts deliberately: the instance is gone, and counting its records there would report the tasks that were just force-stopped.

## Deliberately not in this PR

- **`launchType` filtering** — `TaskRecord` has no launch-type field, so there is nothing to match against. It needs the field added first.
- **`maxResults` / `nextToken`** — no `List*` call in this package paginates. Introducing a token scheme for one would leave the other four inconsistent; it wants doing across all of them at once.
- **`PORTS`/`PORTS_UDP` resources and `networkBindings`**, the other two halves of `mulga-gm5l3`. These are not projection bugs: bridge-mode port publishing is not implemented anywhere. The control plane puts `PortMappings` on the assign payload and nothing in `cmd/ecs-agent` reads it again, `ctrruntime.RunSpec` has no port field, and `NetworkBindings` appears nowhere in the handlers or the agent. Filed separately as `mulga-ic1m8`.

## Testing

`make preflight` passes. Seven new unit tests in `task_accounting_test.go` cover each filter, the combination, the self-exit transition, counts taken from task records, and the regression for the reported symptom — a stale `PlacedTasks` entry no longer inflates the running count.

`TestService_RunTask_PlacesAndAssigns` was asserting `RunningTasksCount == 1` for a task that is PENDING; it now asserts running=0/pending=1. It is not an invariant test.


joshsiv-mulga:
## Live verification on env19

Deployed this branch to env19 (`make cluster-deploy ENV=env19 EXTRA_VARS="spinifex_root=…/spinifex-ecs-accounting"`, exit 0) and exercised the `tfc2-ecs` cluster, which holds both running and stopped tasks.

**Filters are applied, and an omitted `desiredStatus` means RUNNING:**

```
$ aws ecs list-tasks --cluster tfc2-ecs --desired-status RUNNING
  53c020b3-0cd5-4e4d-b30a-544286421b3b
  683df173-34df-443d-920d-b35a824cf1fe
$ aws ecs list-tasks --cluster tfc2-ecs --desired-status STOPPED --query 'length(taskArns)'
4
$ aws ecs list-tasks --cluster tfc2-ecs --query 'length(taskArns)'
2
```

Before this change all three calls returned the same six ARNs.

**A self-exited task leaves the RUNNING filter.** Registered `tfc43-exit:1` (bridge mode, `sh -c "exit 0"`) and ran one task:

```
$ aws ecs describe-tasks --tasks 32954d59-… --query 'tasks[0].{l:lastStatus,d:desiredStatus}'
STOPPED  STOPPED

$ aws ecs list-tasks --cluster tfc2-ecs --started-by tfc43-live --desired-status STOPPED
  32954d59-958a-4e3a-bf07-0ceecc825edf
$ aws ecs list-tasks --cluster tfc2-ecs --started-by tfc43-live
  (empty)
```

Nobody asked that task to stop; it exited on its own and the control plane moved its desired status with it. That also exercises the `startedBy` filter.

**Counts come from the task records.** The cluster's one container instance reports:

```
{"id": "i-b248593927c52f76f", "running": 1, "pending": 0, "status": "ACTIVE"}
```

Two tasks answer the RUNNING desired-status filter but only one is actually running: `53c020b3` has `lastStatus=STOPPED, desiredStatus=RUNNING`. That record predates this deploy, so it is exactly the state the `recordTaskState` change stops being created — and the count is right about it either way, because it is taken from `lastStatus` rather than from `PlacedTasks`.

**Conformance probe TFC-0043 reads `absent`** (was `present`). The probe originally compared the two filters' counts; I have since rewritten it to compare the ARN sets, because a cluster holding as many stopped tasks as running ones would read as broken on a count. Both the count form and the set form give `absent` on this build — the harness change was made after the first verdict, not in order to produce it. Recorded in devlab at `3ac472a` and the follow-up commit.

`TFC-0041` still reads `absent`, so defaulting `desiredStatus` to RUNNING has not starved the probe that lists stopped tasks.

PR #1063, opened 2026-09-16
fix(ec2): project creation-time tags into the central tag store

joshsiv-mulga:
Closes `mulga-bdymy`. Conformance gap TFC-0047.

## What was wrong

EC2 keeps tags in two places: each service's own KV record, read by that resource's describe, and a central S3 index at `tags/{accountID}/{resourceID}.json`, read by `DescribeTags`. `Daemon.createTags` writes both. Creation-time `TagSpecifications` wrote only the record.

So a resource created with tags described as having none, through the tag API, until something tagged it again. `DescribeVpcs` showed the tag; `DescribeTags` did not.

Found by `terraform-aws-modules/eks` 20.x, which reads a cluster security group's tags back after creating it. `aws_ec2_tag` fails against a resource whose tags the tag API cannot see, and the EKS stack got stuck half-applied there.

This is not the `CreateTags` lost-update race in TFC-0009 — I tested that first, and two concurrent `CreateTags` calls both land. It is the write-path split.

## What changed

A `CentralTagWriter` interface in `handlers/ec2/vpc`, implemented by `TagsServiceImpl` with a compile-time check next to the existing `InstanceTagWriter` one. The daemon injects it after constructing the VPC service, alongside `SetDefaultPublicIPMapping` and `SetExternalIPAM`.

`projectRecordTags` is the create-direction counterpart to `updateRecordTags`. `CreateVpc`, `CreateSubnet`, `CreateSecurityGroup` and `CreateNetworkInterface` each call it after their record write. It is a no-op on a nil writer or an empty tag map, and it logs a write failure rather than returning it: the resource exists either way, and failing a create over a tag index would be worse than the divergence it repairs.

## Scope

Four resource types, chosen because they are the ones the EKS module touches and they share one service. Ten others have the same split and are not fixed here: image, volume, snapshot, route table, internet gateway, egress-only internet gateway, elastic IP, NAT gateway, key pair, launch template. They each live with their own service and need the same injection; stated here rather than left for a reader to find.

## Testing

`make preflight` passes. Six unit tests in `central_tags_test.go`: vpc, subnet and security group each project their tags; a create with no tags does not write; a nil writer still creates; and a writer that errors does not fail the create.


joshsiv-mulga:
## Live verification on env19

Deployed this branch to env19 (`make cluster-deploy ENV=env19 EXTRA_VARS="spinifex_root=…/spinifex-tags"`, exit 0), then created one of each covered resource with tags supplied only through `TagSpecifications`, and read them back through `DescribeTags` — the API that could not see them before.

```
vpc=vpc-fa8d42c8c49e7c69b sub=subnet-0ed7b2ccda5ec89e6 sg=sg-6f976158affa022e6 eni=eni-a0e7898594fd648a3

--- describe-tags vpc-fa8d42c8c49e7c69b ---
Name    tfc47-verify
env     live
--- describe-tags subnet-0ed7b2ccda5ec89e6 ---
Name    tfc47-subnet
--- describe-tags sg-6f976158affa022e6 ---
Name    tfc47-sg
--- describe-tags eni-a0e7898594fd648a3 ---
Name    tfc47-eni
```

All four resources were then deleted.

Conformance probe `TFC-0047` reads `absent` (was `present`). The probe requires a second tag written by `CreateTags` to be visible as well, so it cannot read `absent` on an environment where `DescribeTags` is simply broken — that would be a different gap.

The `eks` conformance module, which is where this was found and which previously stopped half-applied at `aws_ec2_tag`, is re-running now; result to follow here.

joshsiv-mulga:
## Correction: the eks module failure is a different gap

I attributed the `eks` module's apply failure to this gap when I filed it. That was wrong, and the from-scratch re-run on this build proves it.

`aws_ec2_tag` uses `CreateTags`, not `TagSpecifications`, so it never touches the creation-time write path this PR fixes. With the fix deployed, the module still fails at the same resource:

```
apply rc=1
Error: reading ec2 resource (sg-71976438fb3c96ede) tag (project): empty result
  with module.eks.aws_ec2_tag.cluster_primary_security_group["project"]
```

The earlier re-run passed only because it resumed a half-applied stack whose security group already existed, so the two `aws_ec2_tag` resources were not created together.

The real cause is a lost update. The central tag store keeps one S3 object per resource and `CreateTags` is a read-modify-write of the whole object, serialised by `TagsServiceImpl.mutex` — a process-local `sync.Mutex`. `ec2.CreateTags` is subscribed on the `spinifex-workers` NATS queue group, so any node may answer any call. Six parallel writers on env19 leave four keys:

```
$ for i in 1 2 3 4 5 6; do aws ec2 create-tags --resources $sg --tags Key=k$i,Value=v$i & done; wait
$ aws ec2 describe-tags --filters Name=resource-id,Values=$sg
k2 v2 / k3 v3 / k4 v4 / k6 v6      # k1 and k5 lost, every call returned success
```

Filed as `mulga-tt77o` (P0) and conformance gap TFC-0048, with a probe that runs a sequential control first and then three parallel trials, so it cannot read `absent` on a single-node environment where the mutex hides the race.

**This PR stands on its own merits.** The creation-time split it fixes is real and independently verified above: a tag supplied through `TagSpecifications` was returned by the resource's own describe and not by `DescribeTags`, on vpc, subnet, security group and network interface. `TFC-0047` moved `present` → `absent` and stays there. It is simply not what was breaking the module.

PR #1064, opened 2026-09-17
fix(ecs): make awsvpc ENI slots a placement resource

joshsiv-mulga:
Closes `mulga-24ujc`. Conformance gap TFC-0046.

## What was wrong

The ENI cap itself is right: `instancetypes.HotPlugENISlotsForType` is `MaxENIsForType - 1`, a t3 gets two task slots, and the daemon allocates distinct QEMU slots for them. Nothing outside the daemon knew the limit existed.

`InstanceRecord` models CPU, memory and GPU, and `fits`/`placeTask` gate on those three. ENIs are not in the model, and the record did not even carry the instance type the cap is derived from. So every over-capacity task was placed successfully, created an ENI, failed to attach it and rolled it back. A service whose desired count exceeds the cluster's slot count churns ENI create/delete forever instead of reporting a placement failure once.

Reproduced on env19 on a single-instance t3.micro cluster: the first two tasks reached RUNNING, the next eight failed identically with

```
{"reason": "RESOURCE:eni", "detail": "attach task ENI eni-08228e1356975a836 -> i-b248593927c52f76f: ServerInternal"}
```

Four divergences in that one line and the behaviour behind it.

## What changed

**ENI slots are a placement dimension.** `InstanceRecord` gains `InstanceType` and `ReservedENIs`. Total slots are derived from the type rather than stored, so a corrected limit table applies on the next read. `fits` and `placeTask` take an ENI count; `RunTask` and `StartTask` pass 1 for an awsvpc task and 0 otherwise; the reservation is committed in the same KV CAS as CPU and memory and released with them.

**An unknown instance type is not gated.** A record written by an agent that predates this change has no type, and refusing every awsvpc placement against it would turn a mixed-version cluster into an outage. A recognised type never yields zero slots (`defaultMaxENIs` is 4), so zero is unambiguously "not reported yet".

**The instance type arrives the way AWS carries it** — the built-in `ecs.instance-type` container-instance attribute, alongside `ecs.availability-zone`. The agent already fetches the IMDS instance identity document for the account ID and that document carries `instanceType`, so this costs no extra IMDS call. Both attributes are projected back on `DescribeContainerInstances`.

**The failure names the resource and nothing else.** `RESOURCE:ENI`, spelled as AWS spells it, and distinct from a plain capacity refusal so a caller can tell which resource bound. The detail is a fixed sentence; the wrapped error keeps `ServerInternal` and the internal ENI and instance IDs in the log. The attach-time path stays as a backstop for a failure the reservation cannot predict, and it gets the same treatment.

**`DescribeInstanceTypes` reports `NetworkInfo.MaximumNetworkInterfaces`** — the string `NetworkInfo` did not previously appear anywhere in the repo, so every type returned null and a caller sizing an ASG had nothing to size against. One helper in `instancetypes`, used by all four generators, reading the same table the slot allocator enforces, so the advertised figure cannot drift from the enforced one. Only fields that are true of this stack are populated: `Ipv4AddressesPerInterface` and the ENA/IPv6 flags are left unset rather than guessed.

**The device index is the slot the device landed on.** `attachENI` stored the requested index verbatim and the daemon ignored it, allocating the real PCIe slot and recording it separately as `HotPlugSlot`. So two task ENIs on one instance both reported device index 1, which cannot happen on AWS. The daemon now writes the allocated slot as the device index.

## Deliberately not done

- **An `ENI` entry under `DescribeContainerInstances.remainingResources`.** AWS does not report one there. Inventing a resource name would be a second divergence dressed as a fix; `NetworkInfo` is where AWS puts this.
- **A device index on the ECS task attachment.** AWS's `Attachment.details` for an `ElasticNetworkInterface` carries `networkInterfaceId`, `macAddress`, `privateIPv4Address` and `subnetId`, and no device index. `taskToAWS` already matches.

## Testing

`make preflight` passes. New tests cover the cap derived from the instance type, an unknown type not being gated, ENI exhaustion as its own refusal distinct from a capacity refusal, the third task declined at placement with **no ENI created for it** (the observable difference from the old churn), the same for `StartTask`, the slot handed back when a task stops, a detail that carries neither `ServerInternal` nor an internal ID, `NetworkInfoForType` agreeing with `MaxENIsForType` and every generated type carrying it, and two ENIs on one instance reporting distinct device indexes.

One existing assertion changed: `TestRunTask_Awsvpc_AttachFailure_RollsBack` pinned `RESOURCE:eni`, the lower-case spelling this corrects. It is not an invariant test.


joshsiv-mulga:
## Live verification on env19

Deployed `feat/ecs-eni-placement` to env19 and rebuilt the container-instance AMI, because the ENI cap is derived from an attribute the **agent** reports and the agent ships inside the guest, not on the host. A control-plane deploy alone leaves every existing instance ungated — which is the behaviour `TestInstanceRecord_UnknownInstanceTypeIsNotENIGated` pins, so a mixed-version cluster degrades to today's behaviour rather than refusing every awsvpc placement.

```
make build-ecs-node-image          # feat/ecs-eni-placement
spx admin images import --ami-name spinifex-ecs-node --tag spinifex:managed-by=ecs
  -> ami-8a7496be843fbbdb8
```

### The instance reports its type

Before, `attributes` was `null`; the scheduler had nothing to derive a cap from.

```
$ aws ecs describe-container-instances --cluster tfc2-ecs --container-instances i-83dfdc842119e726f
attrs:
  ecs.instance-type       t3.small
  ecs.availability-zone   ap-southeast-2a
remainingResources:
  CPU      2048
  MEMORY   1978
```

`remainingResources` still lists CPU and MEMORY only. AWS does not publish an `ENI` entry there either, and the plan records that as deliberately not done.

### DescribeInstanceTypes publishes the cap

Before this change the field was `null` for every type.

```
t3.micro     3
t3.small     3
m5.large     3
m5.4xlarge   8
```

All 62 catalogued types carry `NetworkInfo`, and the figures track `MaxENIsForType` rather than a constant, which is the point of deriving both from one table.

### The third awsvpc task is declined at placement

`t3.small` → 3 ENIs → 2 hot-plug slots. The service holds one, a probe task holds the other. CPU and memory are nowhere near exhausted (2048 / 1978 against a 128/128 task), so nothing but the interfaces can decline this.

```
$ aws ec2 describe-network-interfaces --query 'length(NetworkInterfaces)'
7

$ aws ecs run-task --cluster tfc2-ecs --task-definition tfc46-eni --network-configuration ...
{
    "tasks": null,
    "failures": [
        {
            "reason": "RESOURCE:ENI"
        }
    ]
}

$ aws ec2 describe-network-interfaces --query 'length(NetworkInterfaces)'
7
```

The ENI count is unchanged across the refusal. That is the observable difference from the old behaviour, where each attempt created an ENI, failed to attach it and rolled it back.

### Device indexes are distinct

```
task 09615397…  eni-c7e35b6f975af8883  i-83dfdc842119e726f  deviceIndex=1
task 95dac2ad…  eni-4897144ba2e669e23  i-83dfdc842119e726f  deviceIndex=2
```

Plus `eni-8dae5126154e978d5` at index 0, the instance's own. Previously both task ENIs reported index 1, which cannot happen on AWS.

### The slot comes back

Stopping a task and waiting for the agent to report `STOPPED` lets the next one place:

```
$ aws ecs stop-task --cluster tfc2-ecs --task 3351a730-…
  lastStatus=RUNNING … RUNNING … STOPPED
$ aws ecs run-task …
  tasks: ["arn:…:task/tfc2-ecs/95fda025-b389-4028-96ec-a5043996761a"], failures: null
```

## One code change came out of the verification

`7fdd2abcd` drops the detail from the capacity refusal, so it is now the reason alone.

The conformance probe for this gap asserts `reason == "RESOURCE:ENI" and not detail`, and against the first build it read **`present`** — a correct fix recorded as still broken. The probe is right and the code was wrong: AWS returns `RESOURCE:*` placement failures with no detail, and a sentence that only restates the reason gives a caller a second string to parse for nothing.

I fixed the code rather than relaxing the probe. A probe edited after a verdict is in hand cannot be told apart from one edited to produce that verdict, so the assertion is untouched and both verdicts are on record:

```
probe_TFC_0046   present   # 06819bb89, detail present
probe_TFC_0046   absent    # 7fdd2abcd, reason alone
```

`make preflight` green. Sweep on this build: `absent=27 present=8 skipped=13`, with TFC-0046 on the absent side. TFC-0043 and TFC-0047 read `present` here because this branch predates #1062 and #1063; the EKS skips are the destroyed eks stack, not a regression.

PR #1073, opened 2026-09-17
fix(ec2): clear a resource's central tags when it is deleted

joshsiv-mulga:
Closes `mulga-4gr6s`. Conformance gap TFC-0049.

## What was wrong

`DescribeTags` kept returning a resource's tags after the resource was gone. The central tag store's `DeleteAllTags` had exactly one production caller, `handlers/ec2/instance/service_impl.go:477`, so every resource type other than an instance left its tag entry behind forever.

Found while cleaning up the security group used to reproduce TFC-0048: its six tags were still readable after the group was deleted. Confirmed pre-existing rather than a regression of the tag-store move by finding `eni-a0e7898594fd648a3` in a pre-deploy snapshot long after that interface was gone.

The index therefore grows without bound, a `DescribeTags` with no resource-id filter returns rows the caller cannot describe, and a recycled id inherits the dead resource's tags.

## What changed

- `CentralTagWriter` becomes `CentralTagStore` and gains `DeleteAllTags`. `TagsServiceImpl` already implemented that method and the daemon already injects that value, so this is a rename rather than new wiring.
- `clearRecordTags` is the delete-direction counterpart of the existing `projectRecordTags`: a nil store is a no-op, a failure is logged and swallowed.
- Called from `DeleteVpc`, `DeleteSubnet`, `DeleteSecurityGroup`, `deleteNetworkInterface` and `deleteSecurityGroupInternal`.

Two of those five are not the obvious ones. `deleteNetworkInterface` is shared by the public delete and by `ForceDeleteInstanceENI`, which is what covers an ENI reclaimed by a task or instance going away rather than by an API call — the case that exposed the gap. `deleteSecurityGroupInternal` is the only path the `DeleteVpc` cascade takes to the default security group, so without it the one group a caller never deletes by hand is the one that always leaks.

The clear runs after the record delete has succeeded, never before. Clearing first would strip the tags from a resource that then fails to delete and stays live.

## Deliberately not in this PR

- **The other taggable types** — volumes, snapshots, images, key pairs, route tables, internet and NAT gateways, elastic IPs, launch templates, placement groups. Each owns its delete path in a separate package; they need the same treatment and are kept out so this diff stays reviewable.
- **A reaper for rows already leaked.** Clearing on delete fixes the leak going forward and does nothing for entries already in the index. Nothing in the store can tell a live resource from a dead one without an existence oracle per type, so that is its own piece of work. Not worth building here: these environments are rebuilt rather than kept.
- **A gateway-level hook.** `gateway/ec2.go` has a single action map that would catch every `Delete*` in one place, which is tempting and wrong: it only sees API calls, so it would have missed the ENI case in the bead entirely.

## Testing

`make preflight` passes. A table test drives all four types through the same store, so a type added to this package without a clear call fails there rather than leaking quietly, plus separate tests for the cascaded default SG, a nil store, and a failing store.

Every assertion was proved able to fail. With `clearRecordTags` stubbed to return early:

```
--- FAIL: TestDelete_ClearsCentralTagStore/security_group
        Error: []string(nil) does not contain "sg-e59ece4c7bce56d20"
--- FAIL: TestDelete_ClearsCentralTagStore/subnet
--- FAIL: TestDelete_ClearsCentralTagStore/vpc
--- FAIL: TestDelete_CentralTagStoreError_DoesNotFailDelete
--- FAIL: TestDeleteVpc_ClearsTheCascadedDefaultSecurityGroup
```


joshsiv-mulga:
## Live verification on env19

Deployed this branch (`make cluster-deploy ENV=env19 EXTRA_VARS="spinifex_root=…/spinifex-tag-delete"`, all three nodes `ok=50 changed=8 failed=0`). Nodes confirm the build:

```
spinifex v1.20.0-36-g864f57e98 (864f57e98) linux/amd64
```

**Every type clears, including the one nobody deletes by hand.** Created a tagged VPC, subnet, security group and ENI, tagging the SG through `create-tags` rather than `TagSpecifications` so both write paths are exercised, then tagged the VPC's default security group and deleted everything:

```
--- tags BEFORE delete ---
eni-a07c271af1a38563f       Name      tfc49-eni
sg-677574ca46675c0aa        Name      tfc49-sg
sg-677574ca46675c0aa        purpose   gap-repro
subnet-12b4d0988a28e26e9    Name      tfc49-subnet
vpc-1a216f9c221fd2eef       Name      tfc49-vpc

--- tags AFTER delete (resource-id filter) ---
(empty)

--- unfiltered describe-tags ---
none of the deleted ids appear
```

The unfiltered read matters as much as the filtered one: the acceptance criteria ask for both, and an index that merely stops answering a targeted query would still hand a cost or inventory tool rows for resources that do not exist. `sg-9605e123100f3654a`, the cascaded default group, is gone from the index too — that one is only ever removed by `DeleteVpc`, so it is the case a per-handler fix would miss.

**Probe TFC-0049 reads `absent`** (was `present` on the sweep against `dev c674e4b5c` earlier today, which is the negative control — same probe, same environment, build without this change).

The probe is not trivially passing: it confirms the security group is actually gone via `describe-security-groups` before reading its tags, precisely so a group that failed to delete cannot be read as a fix.

PR #1077, opened 2026-09-17
chore: stop the GitHub schedule firing the E2E Nightly

benduncan:
Removes the `schedule:` block from `e2e-nightly.yml`. Nine lines, no other change.

## Why straight to `main`

#1076 already did this on `dev`, but GitHub registers a `schedule:` only from the **default branch**. So the cron keeps firing until the removal reaches `main`, giving three nightlies a day: the two the Cloudflare Worker dispatches, plus the stale one. All three share the `e2e-hypervisors` concurrency group, so the stale run delays the dispatched ones — run 35257684854 was created 18:15 UTC against a 14:17 cron and did not start until 21:05.

This is a deliberate exception to the sub-repo `dev`-first policy, taken because the thing being fixed is specifically a property of the default branch and cannot be fixed anywhere else.

## Why it is safe on `main`

With the cron gone, a dispatch on `main` gives `GIT_BRANCH = github.ref_name = main`, and every cell's `if:` falls through to `inputs.cells == ''`, which is what selects the full matrix. Blank-cells dispatch running all 21 cells is proven by run 35244484833.

The eleven now-dead `github.event_name == 'schedule'` expressions are left alone here. They are unreachable once nothing fires a schedule, and #1076 already removes them on `dev`, so they arrive with the next merge rather than being deleted twice.

## Merge cleanliness

The replacement comment block is byte-identical to the one on `dev`, verified by diff, so this hunk does not conflict when `dev` next merges to `main`.

## Testing

- `make preflight` passes.
- YAML parses.
- The Worker was proven firing end to end: it dispatched run 35274654136 on `dev` from its own cron at 21:05:09 UTC.


juliansommer:
closed so it doesn't break fast forwarding, its fine.

PR #1079, opened 2026-09-17
feat: report the campaign an install came from

joshsiv-mulga:
## What

`install.mulgadc.com/s/<code>` serves `setup.sh` with `INSTALL_SPINIFEX_CAMPAIGN` assigned, so the installer knows which published link the operator followed. `setup.sh` records it in `/etc/spinifex/campaign`, and `spx admin init` / `spx admin join` send it with the existing call-home.

This is the spinifex half of Cicada phase 1. The other half is in the mulgadc.com repo: the `/r/<code>` redirector, the `mg_src` cookie, the campaign columns on `mulga-events`, and the asset register at `stats.mulgadc.com/assets`.

## Why

Project Cicada is running Reddit, blogs, dev.to crossposts, HackerNews, cold email and a paid Reddit campaign, and nothing today can say which of them produced an install. Stars are the only visible signal and they are not evidence that a channel produces users.

## Safety

The value arrives from the network, so it is accepted only if it matches `^[a-z0-9]{2,4}-[a-z0-9][a-z0-9-]{0,39}$` — checked in `setup.sh` before it is written and again in Go before it is sent. `setup.sh` was driven against a live `install.mulgadc.com` dev worker: a code lands as an assignment directly after the shebang, and `rd-eks; whoami`, `$(id)`, `RD-eks` and a newline-bearing code are all 404s at the Worker and ignored by the script.

A plain `curl install.mulgadc.com | bash` leaves the value empty and the install is reported as unattributed. `SPINIFEX_CAMPAIGN` overrides the file for a node installed by hand. `--no-telemetry` / `SPX_NO_TELEMETRY=1` still sends nothing at all, and the new file is listed in the configuration surface table in `docs/security/network-connections/README.md` so it can be found and deleted.

## Tests

`make preflight` passes. New tests cover the file read, the env override, a malformed override not falling back to the file, and `SendTelemetry` filling the field.


joshsiv-mulga:
Closing: not wanted. Install telemetry already collects enough, and website traffic can come from Cloudflare's own analytics without a new field.

PR #1082, opened 2026-09-18
fix: report the root volume as it was requested

joshsiv-mulga:
Stacked on #1081 — merge that first, and this retargets to `dev` automatically.

## Problem

An instance launches and its root volume works, but almost nothing about that volume reads back the way it was asked for. On env19, a launch naming `/dev/sda1` at 8 GiB with a volume tag returns:

```
DescribeInstances  RootDeviceName                      null
                   RootDeviceType                      null
                   BlockDeviceMappings[0].DeviceName   /dev/vda
DescribeVolumes    Tags                                []
                   Attachments[0].Device               ""
```

## Change

Three of those are omissions with nothing on the other side, and this fixes them.

`RootDeviceName` and `RootDeviceType` are projected onto the instance from `ebsmetadata.RootDeviceName` and `ebs`. `DescribeImages` already reported both for the same AMI, so the two calls disagreed about one image.

The root volume's `EBSRequest` now carries the device name `parseVolumeParams` resolved. `markAttachedVolumesInUse` passes it to `UpdateVolumeState`, which writes it to the volume's metadata, which is what `DescribeVolumes` reports as `Attachments[].Device` — previously an empty string on an attached volume.

Volume-scoped `TagSpecifications` are applied to the root volume's metadata document. Only instance-scoped tags were read, and the comment claimed volume specs were "handled elsewhere"; nothing handled them anywhere.

## What this deliberately does not change

`BlockDeviceMappings[].DeviceName` still reports the guest path. `UpdateGuestDeviceNames` rewrites it on every launch and start, and the comment at `vm/volumes.go:254` states the intent: a caller who SSHes in and runs `lsblk` should see names matching `DescribeInstances`. That is a deliberate convention recorded in code and in no spec, and it is also an AWS divergence with real cost — `aws_instance.ebs_block_device` is ForceNew, so a device name that never reads back as sent replaces the instance on every plan. Reconciling the two is a decision, not a bug fix, so it is reported and left alone. IMDS `block-device-mapping` reads the same field and would move with it.

With this merged the instance still replaces on every plan, because the device name alone is enough to force it. Worth saying plainly rather than claiming a fix that does not move that symptom.

`Encrypted` reporting `true` against a request for `false` is also left: the platform encrypts every volume at rest when a master key is configured, so `true` is the honest answer about the volume that exists, and AWS reports actual state the same way for an account with encryption by default.

## Verification

PLACEHOLDER

Part of `mulga-mdafn`.


joshsiv-mulga:
Verified live on env19 after `make cluster-deploy` from this branch. The same launch as in the description — `/dev/sda1` at 8 GiB with `ResourceType=volume` tag `Name=tfc0056-root` — now reads:

```
DescribeInstances  RootDeviceName                      /dev/sda1
                   RootDeviceType                      ebs
                   BlockDeviceMappings[0].DeviceName   /dev/vda      (unchanged, deliberate)
DescribeVolumes    Tags                                [{Name: tfc0056-root}]
                   Attachments[0].Device               /dev/sda1
```

The divergence for the remaining `/dev/vda` reading is recorded as D-13 in mulgadc/mulga#142.

PR #1098, opened 2026-09-21
fix(gpu): support multi-GPU instances end to end

konovenski:
## Summary

- claim the GPU count advertised by an instance type, with an optional `Type=gpu` launch override for homogeneous local pools
- preserve every whole-GPU or MIG attachment through launch, stop/start, the admin API, and the web UI
- render complete GPU lists on the Nodes table and instance detail page

## Changes

- make multi-device claims all-or-nothing and gate instance capacity by free GPU count
- change the admin VM contract from singular `gpu` to `gpus`, resolving every stored attachment
- add multi-GPU backend and component coverage, rebuild the embedded frontend, and document multi-GPU launches

## Testing

- `go test ./spinifex/handlers/ec2/instance`
- `go test ./spinifex/daemon -run 'TestDaemonGPUClaimer|TestHandleNodeVMs'`
- `pnpm exec vitest run 'src/routes/_auth/ec2/-components/gpu-detail-card.test.tsx'`
- `pnpm lint`
- `pnpm build`
- Linux/amd64 `spx` cross-build
- live 5-GPU node smoke test: the admin API returned two attachments for one running instance and three for the other; the Nodes table rendered both lists and the three-GPU detail page rendered `GPUs (3)` with all PCI addresses


konovenski:
ola my friends, this PR was vibe-coded. in my own words - i have 8 gpus and when migrating from proxmox it was a little surprising that i can't assign more than one gpu to an instance. This pr is basically making it an attribute that you can add into aws cli command and assign arbitrary gpu count. it is very useful when you're doing multi-gpu nccl workflows and so on.

I'll be ok if you will close that pr and reimplement it yourself, but i submitted this one as basically something i requested astra to implement, run test workflows, change UI, etc. and then on top i validated it myself. It is working and is ready to merge/reimplement/etc

juliansommer:
Hey, thanks for the contribution and for being upfront about how it was built. 

Most of this is correct and I'll end up keeping it. The one piece we can't take is the Type=gpu count override. In EC2 the GPU count is a fixed property of the instance type, there's no API to request N GPUs. ElasticInferenceAccelerators is a real EC2 parameter for Elastic Inference accelerators, so real EC2 rejects Type=gpu with InvalidParameterValue. We keep the EC2 surface compatible, so we can't overload that field.

Currently any GPU we don't recognise falls back to the g5 family, and every g5 size is single-GPU, so the count was pinned to 1. The fix is to put the real AWS multi-GPU sizes in the catalogue and match your hardware to them. On your 8-GPU box that means g5.48xlarge-class types for using all 8, plus 4 and 2 GPU sizes for smaller jobs. These are all standard EC2 type names, so RunInstances stays identical to AWS.

Arbitrary GPU counts such as 3 won't be covered initially since AWS has no 3 GPU type. We may followup later and allow custom operator defined types pending interest (since it is a direct AWS divergence).

This will ship soon. Thankyou for using Spinifex and let me know if you have any more feature requests :)

